Issue #3, opened 2015-03-21
disable pex and dht


jzelinskie:
there's a cli flag for disabling trackers; it would be super useful to get some to disable pex and dht

jzelinskie:
It looks like there's already an option for disabling DHT in `torrent.Config`, so it's just the work of adding the flag.

anacrolix:
Will do. It would be nice to use something that exposes config structs as flag.FlagSets. All of the torrent-related binaries want to expose the same set of flags.

PR #5, opened 2015-03-22
add cli flag to disable DHT

jzelinskie:
This partially addresses #3.


cmd/torrent/main.go
```diff
@@ -43,6 +44,7 @@ func main() {
 		DisableTrackers: *disableTrackers,
 		ListenAddr:      *listenAddr,
 		NoUpload:        !*upload,
+		NoDHT:           *disableDHT,
```
anacrolix: Isn't this back to front? I think NoDHT: *disableDHT?

PR #10, opened 2015-04-02
Added BootstrapNodes to dht.ServerConfig

milosgajdos83:
This is beneficial to users who are running their own private DHT servers or have outbound connections to port 6881 disallowed on firewalls.


anacrolix:
Great work

PR #12, opened 2015-04-11
Speculative PR :-) Making using DHT more idiomatic Go

milosgajdos83:
NewServer returns reference to Server object but does not start
listening automatically. This is done via public Serve() method.
This makes writing server code more idiomatic Go - I have modified
dht-server code in cmd directory. ie. now the API now allows you to write code like
log.Fatal(s.Serve()) as oppose to empty selects and infinite loops etc.
I have also updated the tests which reflect the new API.
You will probably reject this PR but I'm giving this a shot anyway :)

### DIFF
--- a/client.go
+++ b/client.go
@@ -588,6 +588,12 @@ func NewClient(cfg *Config) (cl *Client, err error) {
 		if err != nil {
 			return
 		}
+		go func() {
+			err := cl.dHT.Serve()
+			if err != nil {
+				panic(err)
+			}
+		}()
 	}
 
 	return
@@ -614,6 +620,9 @@ func (me *Client) Close() {
 	me.event.Broadcast()
 	for _, t := range me.torrents {
 		t.close()
+	}
+	if me.dHT != nil {
+		me.dHT.Close()
 	}
 }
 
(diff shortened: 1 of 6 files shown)


anacrolix:
I've seen this pattern before, could you link me to something suggesting this is the canonical way to do things? Additionally, I think a .Serve call is now required in torrent.NewClient, could you add that?

milosgajdos83:
I'll try to show you some examples, but feel free to ignore me - I'm still a pretty rookie gopher, so...it was just my observation based on other code I've seen not only in standard library.

One of the examples is in the standard library in http package.
[Here](http://golang.org/src/net/http/server.go?s=53504:53559#L1796) is a ```ListenAndServe``` which returns ```error``` which allows you to write code like [this](http://golang.org/src/net/http/doc.go#L78). 

If you dig in further into the source you can see that the private ```serve()``` [method](http://golang.org/src/net/http/server.go#L1124) is called in a separate goroutine from within the public ```Serve()``` method [here](http://golang.org/src/net/http/server.go#L1751).

anacrolix:
It's an interesting change, but I don't feel it improves anything. It's possible in the future there'd be a reason to have Serve separately, but for now that's not the case, so I'm loath to change it for the sake of it. Thank you for the effort.

PR #14, opened 2015-05-16
Multiple file metainfo mode support in cmd tools

milosgajdos83:
```torrent-create``` and ```torrent-verify``` did not work correctly when a single file is present in ```metainfo``` torrent file. ```torrent-create``` now correctly fails if the path passed as an command line argument does not exist - otherwise it throws runtime panic as the non-existent file case is not handled by it. ```torrent-verify``` now has a summary flag. Finally, in ```span.go``` I've cheekily sneaked in a change which defines the ```span``` type more like idiomatic Go (I have a weird OCD :-)).

### DIFF
--- a/cmd/torrent-create/main.go
+++ b/cmd/torrent-create/main.go
@@ -26,6 +26,9 @@ func init() {
 func main() {
 	b := torrent.Builder{}
 	for _, filename := range flag.Args() {
+		if _, err := os.Stat(filename); os.IsNotExist(err) {
+			log.Fatal(err)
+		}
 		if err := filepath.Walk(filename, func(path string, info os.FileInfo, err error) error {
 			log.Print(path)
 			if info.IsDir() {
--- a/cmd/torrent-create/main.go
+++ b/cmd/torrent-create/main.go
@@ -26,10 +26,10 @@ func init() {
 func main() {
 	b := torrent.Builder{}
 	for _, filename := range flag.Args() {
-		if _, err := os.Stat(filename); os.IsNotExist(err) {
-			log.Fatal(err)
-		}
 		if err := filepath.Walk(filename, func(path string, info os.FileInfo, err error) error {
+			if _, err := os.Stat(path); os.IsNotExist(err) {
+				return err
+			}
 			log.Print(path)
 			if info.IsDir() {
 				return nil
--- a/cmd/torrent-verify/main.go
+++ b/cmd/torrent-verify/main.go
@@ -16,16 +16,40 @@ import (
 )
 
 var (
-	filePath = flag.String("torrent", "/path/to/the.torrent", "path of the torrent file")
-	dirPath  = flag.String("path", "/torrent/data", "path of the torrent data")
+	torrentPath = flag.String("torrent", "/path/to/the.torrent", "path of the torrent file")
+	dataPath    = flag.String("path", "/torrent/data", "path of the torrent data")
+	summary     = flag.Bool("summary", false, "display summary at the end")
 )
 
-func init() {
-	flag.Parse()
+func verifySummary(sMap map[bool][]int) {
+	fmt.Println("----------------")
+	fmt.Println(" TORRENT-VERIFY ")
+	fmt.Println("----------------")
+	fmt.Printf("Number of correct pieces: %d\n", len(sMap[true]))
+	fmt.Printf("Number of wrong pieces: %d\n", len(sMap[false]))
+}
+
+func fileToMmap(filename string, length int64, devZero *os.File, mMapSpan *mmap_span.MMapSpan) {
+	osFile, err := os.Open(filename)
+	if err != nil {
+		log.Fatal(err)
+	}
+	mmapFd := osFile.Fd()
+	goMMap, err := gommap.MapRegion(mmapFd, 0, length, gommap.PROT_READ, gommap.MAP_PRIVATE)
+	if err != nil {
+		log.Fatal(err)
+	}
+	if int64(len(goMMap)) != length {
+		log.Printf("file mmap has wrong size: %#v", filename)
+	}
+	osFile.Close()
+	mMapSpan.Append(goMMap)
 }
 
 func main() {
-	metaInfo, err := metainfo.LoadFromFile(*filePath)
+	flag.Parse()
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: cmd/torrent-create/main.go [hunk 1]
```diff
@@ -26,6 +26,9 @@ func init() {
 func main() {
 	b := torrent.Builder{}
 	for _, filename := range flag.Args() {
+		if _, err := os.Stat(filename); os.IsNotExist(err) {
+			log.Fatal(err)
+		}
 		if err := filepath.Walk(filename, func(path string, info os.FileInfo, err error) error {
```
anacrolix: I think you've introduced a race via LBYL. You should check the error inside the Walk? This is the reason for the panic.

cmd/torrent-verify/main.go
```diff
@@ -16,16 +16,40 @@ import (
...
 )
 
-func init() {
-	flag.Parse()
+func verifySummary(sMap map[bool][]int) {
+	fmt.Println("----------------")
+	fmt.Println(" TORRENT-VERIFY ")
+	fmt.Println("----------------")
+	fmt.Printf("Number of correct pieces: %d\n", len(sMap[true]))
+	fmt.Printf("Number of wrong pieces: %d\n", len(sMap[false]))
```
anacrolix: I think you could just blow away what's currently logged, and show a progress line with something like: X% complete (X/Y pieces correct). When you finish, print a newline?

cmd/torrent-verify/main.go
```diff
@@ -16,16 +16,40 @@ import (
...
+	mmapFd := osFile.Fd()
+	goMMap, err := gommap.MapRegion(mmapFd, 0, length, gommap.PROT_READ, gommap.MAP_PRIVATE)
+	if err != nil {
+		log.Fatal(err)
+	}
+	if int64(len(goMMap)) != length {
+		log.Printf("file mmap has wrong size: %#v", filename)
+	}
+	osFile.Close()
+	mMapSpan.Append(goMMap)
```
anacrolix: Simpler to just return the gommap.MMap, and append it in the caller.

dht/dht.go
```diff
@@ -980,11 +980,13 @@ func bootstrapAddrs(nodeAddrs []string) (addrs []*net.UDPAddr, err error) {
 		}
 	}
 	for _, addrStr := range bootstrapNodes {
-		udpAddr, err := net.ResolveUDPAddr("udp4", addrStr)
-		if err != nil {
-			continue
+		if addrStr != "" {
```
anacrolix: I don't have a name for this, but it's like passing the buck on a sentinel value. You've created a new special case which is addrStr=="". Maybe it could mean something. Maybe not. It's not our problem unless we care about the errors we're getting, which we don't. What do you think of that?

mmap_span/span.go
```diff
@@ -4,9 +4,7 @@ type sizer interface {
 	Size() int64
 }
 
-type (
-	span []sizer
-)
+type span []sizer
```
anacrolix: Nice catch.

cmd/torrent-verify/main.go
```diff
@@ -16,16 +16,40 @@ import (
 )
 
 var (
-	filePath = flag.String("torrent", "/path/to/the.torrent", "path of the torrent file")
-	dirPath  = flag.String("path", "/torrent/data", "path of the torrent data")
+	torrentPath = flag.String("torrent", "/path/to/the.torrent", "path of the torrent file")
+	dataPath    = flag.String("path", "/torrent/data", "path of the torrent data")
```
anacrolix: Currently torrent-verify works by verifying the data in the current working directory. The OS already has a provision for setting the working directory, I think adding another flag just complicates the the cmd's interface. A doc note saying "data is read from the working directory" could be better.

cmd/torrent-create/main.go
milosgajdos83: I tried to avoid the ```Walk``` completely if the ```root``` path does not exist. I'll move this to the ```Walk``` now.

cmd/torrent-verify/main.go
milosgajdos83: The reason why I added the summary flag is, if you have a file with thousands of pieces they're logged in the ```stdout``` and unless you parse them with some UNIX cli tool magic it can get pretty verbose. 
The new ```summary``` flag is optional and just gives a quick overview only if you wish to see one. I'm happy to remove it completely.

cmd/torrent-verify/main.go
milosgajdos83: Yep, great suggestion. That way the function name would also be more correct semantically :+1:

cmd/torrent-verify/main.go
anacrolix: I'm suggesting you ditch the old way, and just go with this. But print a nice progress bar as it goes. Are you familiar with the carriage return character? I did a crude form of it in cmd/torrent: https://github.com/anacrolix/torrent/blob/master/cmd/torrent/main.go#L150

anacrolix:
I appreciate the improvements, keep them coming, thanks.

dht/dht.go
milosgajdos83: I agree with this. It is an ugly hack :-) The reason why I have added the check is, when you are starting your own bittorrent DHT and want to avoid the external ones (ie. utorrent.com and bittorrent.com), currently the ```bootstrapNodes``` must have at least one item.

I need to think about a way to start the "private/internal" bittorrent DHT in some cleaner way than this terrible hack :+1: 
Do you have any suggestions here ?

cmd/torrent-verify/main.go
milosgajdos83: I kinda like the original idea of having a visibility of which piece is incorrect.  I'm happy to implement the progress bit, but like I said we will lose the per-piece visibility, which does not necessarily have to be a bad thing ?

milosgajdos83:
I've addressed some of your comments. I think we can leave the summary bit out. It does not really improve much to be honest. I'd like to keep the path to data flag called ```dataPath``` instead of ```dirPath```. I need to think about that DHT stuff - I've removed my ugly hack :-) 
Thanks for the proper code reviews!

anacrolix:
Great work.

PR #16, opened 2015-06-01
torrent-pick tool

scr4t:
Picks only one file from torrent.


anacrolix:
Bit messy, but gets the job done. Cheers

PR #31, opened 2015-10-10
fix: added announce to torrent spec

kupsef:
The announce attribute was ignored while creating spec from metaInfo. Because of that torrents without announceList weren't announced to the tracker, rendering them useless.


anacrolix:
Good catch, thanks.

Issue #35, opened 2015-10-27
validateInfo in misc.go is offbyone


soul9:
ValidateInfo should work perfectly fine except for a -1 that slipped in the length sanity verification somehow here https://github.com/anacrolix/torrent/blob/master/misc.go#L119

anacrolix:
That's for rounding up integer division. Do you have an example or test case for broken behaviour you're experiencing?

soul9:
sure, here is the testcase
https://gist.github.com/soul9/b2e8cdd3f069222d0e65

anacrolix:
Thanks for that. The error is actually in the builder code, which I incorporated from another project. metainfo/builder.go:307. I'll issue a fix soon.

PR #36, opened 2015-10-27
fix https://github.com/anacrolix/torrent/issues/35

soul9:
fixes an offbyone issue which prevents from using torrents created by torrent library itself


anacrolix:
This was an error in the builder, this fix only works around that bug by also being incorrect. Thanks for raising the issue.

PR #42, opened 2015-12-05
Nicer godoc strings for dht package

dborzov:
Hey! I added some docstrings to the `dht` package. I attempted to stick to the official godoc conventions as described in [here](http://blog.golang.org/godoc-documenting-go-code).

I also copied some lines from the [spec](http://www.bittorrent.org/beps/bep_0005.html). That should not be an issue as the BEPs are public domain.

Let me know what you think. 

Thanks!


anacrolix:
Great! I'll do and checks later and merge.

dborzov:
I noticed that dht.go is quite long and moved some definitions (for `Server`, `Transaction` & `nodeInfo`) to separate dedicated files to offload it. Would you agree with this?

PR #43, opened 2015-12-07
Add usage message to dht-ping cli

dborzov:
This: 
* prints a nice message on how to use dht-ping cli tool when appropriate (when called without arguments, invalid arguments& flags passed,  called with the `-help` key)
* Fixes package docstring for dht package


anacrolix:
I've cherry-picked some of your patches, and rewritten dht-ping.

PR #44, opened 2015-12-11
Implemented Incoming-Query Hook System, Permit Manual IDs.

cathalgarvey:
Hey, it looks like these can't be merged for some reason I'll look into right away, but given the pace of development I thought it best to open the PR now and make it clear what I'm up to. :)

So, I've changed a few interfaces to expose some of the internals. The overarching goal is that I'd like `torrent` to be more useful for creating DHT-based applications that don't actually care about bittorrent per-se. For example, an open source DHT search engine, or a DHT secure message app, etcetera. For that, sometimes the ability to set ones own ID is important, but the ability to define custom event hooks is more important.

For example, if I wanted to implement a DHT search engine, the usual way is to set up lots of cheap nodes that listen for "find_node" queries and investigate any infohashes they harvest for active torrents. Whether one likes this practice or not, there are several large bittorrent search engines based on this practice and no open source implementations I'm aware of. With a hook system, this is trivial to set up in `torrent`.

Or, to implement an open torrent-health API for trackerless torrents, likewise you'd want to have a hook for new peers contacting you in a given swarm. Again; simple enough to set up.

Anyways, something's not auto-mergeable so I'll go digging and report back here. Let me know if you like the general approach or not. I added a test case for the new hook system but it's sparse so far, and needs more coverage. I also made a few little changes here and there, nothing serious; things like using testify more among the existing tests.


cathalgarvey:
Christ on a bike, things changed since I pulled on Nov 22nd. :)

My solution is to take a more recent commit and copy/paste in the meat of the changes I was making. Expect a pretty different set of commits that should bring this in line, shortly.

dht/addr.go
```diff
@@ -7,7 +7,7 @@ import (
 )
 
 // Used internally to refer to node network addresses.
-type dHTAddr interface {
+type DHTAddr interface {
```
anacrolix: Rather than expose this performance hack type, when you pass addrs out of dht, just pass the *udp.Addr as a net.Addr.

dht/dht.go
```diff
@@ -49,24 +49,39 @@ type Server struct {
...
+	Conn   net.PacketConn
 	// Don't respond to queries from other nodes.
 	Passive bool
 	// DHT Bootstrap nodes
 	BootstrapNodes []string
+	// Disable bootstrapping from global servers even if given no BootstrapNodes.
+	// This creates a solitary node that awaits other nodes; it's only useful if
+	// you're creating your own DHT and want to avoid accidental crossover, without
+	// spoofing a bootstrap node and filling your logs with connection errors.
+	NoBootstrap bool
```
anacrolix: Looks good, I assume I use the default bootstrap nodes if none are provided in BootstrapNodes, which is bad style.

dht/dht.go
```diff
@@ -75,6 +90,12 @@ type ServerConfig struct {
 	IPBlocklist iplist.Ranger
 	// Used to secure the server's ID. Defaults to the Conn's LocalAddr().
 	PublicIP net.IP
+	// Hook functions to be called when matching KRPC queries arrive.
+	// Hooks run prior to default KRPC handling, and may modify messages
+	// or indicate whether to skip default handling.
+	// Hooks can therefore replace, extend, or disable KRPC query handling
+	// in a Query-type specific way.
+	KRPCHooks map[string]KRPCHook
```
anacrolix: I'd rather you just exposed a function "OnQuery".

dht/dht.go
```diff
@@ -141,6 +163,12 @@ func NewServer(c *ServerConfig) (s *Server, err error) {
 		}
 	}
 	s.bootstrapNodes = c.BootstrapNodes
+	if c.NodeId != "" {
+		if !c.NoSecurity {
+			return nil, ErrConflictingConfigNoSecNodeId
```
anacrolix: I don't see why this is true. If security is enabled, can't we just check if their provided NodeId is secure, and only error if it isn't?

dht/dht.go
```diff
@@ -156,16 +184,18 @@ func NewServer(c *ServerConfig) (s *Server, err error) {
...
 		}
 	}()
-	go func() {
-		err := s.bootstrap()
-		if err != nil {
-			select {
-			case <-s.closed:
-			default:
-				log.Printf("error bootstrapping DHT: %s", err)
+	if !s.config.NoBootstrap {
```
anacrolix: Better here to actually move the smarts about NoBootstrap into the bootstrap function. By "turning off" sections of code on special conditions, you create more opportunity for bugs to be missed by creating conditionally dead code. I hope this makes sense.

dht/dht.go
```diff
@@ -618,13 +648,28 @@ func (s *Server) nodeByID(id string) *node {
...
+		newmsg *Msg
+		skip   bool
+	)
 	node := s.getNode(source, m.SenderID())
 	node.lastGotQuery = time.Now()
 	// Don't respond.
 	if s.config.Passive {
 		return
 	}
+	// Call any hooks that may apply
```
anacrolix: I think the OnQuery hook should be called even if the server is set to passive. Move this up above the passive check?

dht/dht.go
```diff
@@ -618,13 +648,28 @@ func (s *Server) nodeByID(id string) *node {
...
+	if hook, ok := s.hooks[m.Q]; ok {
+		newmsg, skip = hook(&source, node, &m)
+		if newmsg != nil {
+			m = *newmsg
+		}
+	}
+	// Unless told not to by hook, proceed with normal handling.
+	if skip {
```
anacrolix: "propagate" might be a better name here.

dht/dht.go
```diff
@@ -618,13 +648,28 @@ func (s *Server) nodeByID(id string) *node {
...
 	node := s.getNode(source, m.SenderID())
 	node.lastGotQuery = time.Now()
 	// Don't respond.
 	if s.config.Passive {
 		return
 	}
+	// Call any hooks that may apply
+	if hook, ok := s.hooks[m.Q]; ok {
+		newmsg, skip = hook(&source, node, &m)
+		if newmsg != nil {
```
anacrolix: I think you're better off allowing the callee to modify the Msg in place, and telling them that any changes they make will be propagated.

dht/dht.go
```diff
@@ -208,8 +238,8 @@ func (nid *nodeID) ByteString() string {
 	return string(buf[:])
 }
 
-type node struct {
-	addr          dHTAddr
+type Node struct {
```
anacrolix: I'm not sure there's any value to exposing this at all. The OnQuery user should have no business or interest in real-time updates on the node internals. Nothing is exposed on it. I think it's sufficient just to pass out the node's "source" addr when queries come in, the ID is provided in the msg. Certainly I don't think the external user should be maintaining references to node, as they are dropped and replaced from the node table beyond the control of the user.

anacrolix:
Yeah sorry about that, @dborzov had a big clean up. It's an interesting idea, I look forward to your commit. It's worth noting that the query handling system that's currently there is pretty poor. Your change appears to essentially boil down to wanting to act on incoming queries. I've made some extensive code comments, I hope they're useful.

dht/dht.go
cathalgarvey: Do you mean that users should add a single function that handles any given thing they want, using something like the KRPCHook interface, and then indicates wholesale whether to continue with processing?

I could certainly change it to a single-function interface, though I feel like the map interface is easier to reason about and compose, for complex projects. It elides the redundant task of determining and dispatching the query by name, and discourages any potentially weird handler crossovers. But, if people want weird crossovers they can add KRPCHook-compliant methods of some custom struct/subrouter as easily as sole functions.

Still, it's your party; if you'd rather a single interface function I can refactor. Do you like the KRPCHook interface overall? Will comment on the suggestion of edit-msg-in-place vs. return-new-msg on that line comment.

dht/dht.go
cathalgarvey: Very true, not sure how I got that mental block. Probably I was only thinking about my use-case, which is "search engine / DHT-crawler" where hosting isn't on the cards anyway, and you just keep picking arbitrary IDs without trying to 'fit in'. Your suggestion of keeping these separate and using the docs to point out the connection is far more sensible overall.

dht/dht.go
cathalgarvey: I will confess that, while I think disabling global router bootstrap is a useful option overall, I added it in haste because I wanted to work on test cases and wanted a way to make an isolated DHT on my PC without risk of connecting to the DHT and earning a bad name. :)

So yes, placing this logic somewhere more..logical would be good, as would making what it means more explicit, such as "NoGlobalBootstrap" or something.

dht/dht.go
cathalgarvey: I thought about this; it's a question of option primacy, but also being consistent with what the options "promise". That is, "Passive" means "don't respond", but hooks could hypothetically respond, so that breaks what looks like a guarantee.

Keeping that guarantee would have required one of:

* Overriding hooks brutally (which is what I opted for, for simplicity)
* Changing response logic in a more dramatic way (so that if either routing logic or hooks called a response method, it would check and block there instead), which I felt would be a pretty big PR to issue first-time. :)

..or, discard the guarantee, and let hooks handle things before Passivity is checked. In this case, though I feel like "Passive" should be renamed to "DisableDefaultResponse" or something, which would make it clearer that it will still allow non-default responses through Hooks, while disabling what comes after.

dht/dht.go
cathalgarvey: Acknowledged; so change the interface of whatever hook system we use to return `propagate bool` and check `if !propagate` instead.

dht/dht.go
cathalgarvey: I think this is a "big picture" decision; or, performance vs. stability.

My instinct was stability; to return a new msg, because that avoids any potential side-effects if other parts of router logic log or depend on incoming messages remaining true to their state upon arrival. Message variables in the Query handlers would get overwritten but the original value should remain the same (I think?). That's a nice guarantee to have when developing the router, especially for logging, reports, and perhaps even for the blacklist. However, it comes at the cost of allocating new messages.

If the point of `torrent` is to be highly performant and/or to minimise memory allocations / GC cycles, then editing in place makes more sense, but it may lead to side-effects or even race conditions if messages are currently, or in the future, expected to be static.

Again; your party, your call. :)

dht/dht.go
cathalgarvey: Hm, the inconsistency is because I exposed Node but neglected (yet) to expose its attributes. My reasoning again is from use-case; passive monitoring nodes, or active search nodes, etcetera. So, `Node.Id` (currently `Node.id`) is something I'd like hooks to have access to, as well as `Node.Addr`. I'd even expose `Node.AnnounceToken` for informational purposes, but not the `last..` bits.

The reasoning is, there are no good 'research tool' libraries for DHT analysis, so even if I can't think why someone might want to know about a particular datum, unless it's fairly important for it to remain private I feel like making it public would make `torrent` more useful and distinctive. I initially tried to make changes of this kind to `github.com/nictuku/dht` but would have had to refactor the entire thing messily to add those features. The modular nature of `torrent` makes it much easier to add DHT-only features without breaking the rest of it.

Without making Node public, I could find other ways to expose information though; for example, by attaching methods to Msg that allow it to fetch Msg-requester information, or by creating a special struct for "QueryHandler" information and populating it only when a hook is found from private information known to the server, then passing it in. That might allow one to expose information useful to hook users without fixing a `Node` public interface that might impose a maintenance burden, if that's your concern?

cathalgarvey:
Hey @anacrolix, thanks for the extensive feedback. I've replied inline, and I agree with a lot of your comments right away. Others, I think merit discussion on the intended outcome, and I can go either way depending on the style or philosophy you'd prefer. The only one I mildly disagree with is refactoring hooks to use a single `OnQuery` function, because I feel it'll introduce redundant dispatching and encourage potentially 'naughty' code, but that's a stylistic difference; I can refactor if you insist. :)

I'm in the process of applying the changes I'd made previously to the newly refactored code from yesterday's last commit, and when I'm finished making that consistent with my poor confused git logs, I'll push it up and it should join this PR. Hopefully then it'll be merge-compatible again.

I'm glad to see the @dborzov refactors, even if they made this PR more complicated. Those structs needed a new home. :)

cathalgarvey:
OK, that's brought the changes as-discussed-so-far in line with Master. Now I can make stylistic or logical changes according to the above. :)

dht/addr.go
cathalgarvey: Fixed! Now casts to net.Addr prior to passing to hooks. I'm not as familiar with the *.Addr types as you, so I'll take it on faith for now that casting is safe. :)

dht/dht.go
cathalgarvey: Fixed, and further added some logic so that ID is given to the config as a hex string rather than requiring caller to make a bytestring and cast to string.

dht/dht.go
cathalgarvey: Fixed, changed to propagate. Lead to a stupid bug where I forgot to set propagate to "true" by default, so the Ping test broke. :)

dht/dht.go
cathalgarvey: Still looking into this while I resolve the rest. Without creating a new object with a guaranteed interface that carries "data hooks can use", exposing this and its attributes is the cleanest way I can see to give hook authors access to useful data?

cathalgarvey:
Hey, just wondering if there's still interest in this. I fixed everything I felt was agreed upon, I'm happy to defer to you on the rest but I had some concerns or suggestions about implementation. Looking forward to getting the last bits fixed so I can look at working on other areas of code; would like to get the builtin KRPC stuff working. :)

anacrolix:
Sorry yes, I've just been buried at work. I'll get to it in the next few days.

dht/server.go
```diff
@@ -331,6 +359,9 @@ func (s *Server) getNode(addr dHTAddr, id string) (n *node) {
 	if len(s.nodes) >= maxNodes {
 		return
 	}
+	// (@cathalgarvey) DHT Security is supposed to prevent insecure nodes from
+	// hosting data: not to exclude them from DHT routing entirely. Does
+	// this check exclude them entirely from the local routing table?
```
anacrolix: Yes it does. You can put a comment here saying that's the intended effect for the next guy if you like.

dht/dht.go
```diff
@@ -52,6 +66,8 @@ type ServerConfig struct {
 	IPBlocklist iplist.Ranger
 	// Used to secure the server's ID. Defaults to the Conn's LocalAddr().
 	PublicIP net.IP
+	// Hooks for the server to call on various KRPC queries
+	KRPCHooks map[string]KRPCHook
```
anacrolix: I feel this doesn't reflect what it actually does. Further, the "hook" function can work out what the query is for itself, since it receive the message. I think this should simply be OnQuery func(*Msg, net.Addr) bool. With a note that the Msg will be processed with any changes made if the function returns true. It's much clearer.

dht/dht.go
```diff
@@ -112,7 +128,7 @@ func (nid *nodeID) ByteString() string {
 	return string(buf[:])
 }
 
-type node struct {
+type Node struct {
```
anacrolix: This should not be made public. A user of OnQuery (and presumably other future hooks) can build their own own node table if they want. Nodes are transient and should not be visible. gofmt -r 'Node -> node' -w will make this easy to fix, or try gorename.

dht/krpc_hook.go
```diff
@@ -0,0 +1,14 @@
...
+// A hook function for handling incoming KRPC messages.
+// Hooks can't pass errors back, so if error logging or
+// management within hooks is desired, consider passing
+// methods of a custom type as KRPCHooks and use the type
+// to handle errors.
+// The arguments are the incoming DHTAddr and Msg.
+// The outputs are:
+//  - Optional replacement Msg for ensuing use in default handlers
+//  - Bool indicating whether to propagate msg to default handlers.
+type KRPCHook func(*net.Addr, *Node, *Msg) (*Msg, bool)
```
anacrolix: I think this file should be removed. OnQuery func(*Msg, net.Addr) bool is fine. Note that net.Addr is an interface type, so you don't need to make it a pointer.

dht/server.go
```diff
@@ -458,6 +489,11 @@ func (s *Server) announcePeer(node dHTAddr, infoHash string, port int, token str
 	if port == 0 && !impliedPort {
 		return errors.New("nothing to announce")
 	}
+	// Get node, verify that it's secure for storing announces on.
+	N := s.getNode(node, "") // May not exist, probably prone to races
```
anacrolix: Yeah I don't think this is right. getNode wasn't intended to be used this way. Furthermore, the decision regarding whether to announce to a given peer should be made outside the announcePeer function. I'm not there's any value to checking whether the node we're announcing to is "secure". Security here is used to ensure that messages we're *receiving* cannot be from sources with arbitrary IDs, allowing those nodes to easily corrupt the data stored on the network.

anacrolix:
Thanks Cathal, I've cherry-picked parts of your PR into cleaner commits and applied the more opinionated ideas I've listed above. I hope this sits well with you, and that you can proceed with your use case with these changes.

cathalgarvey:
Thanks! Delighted to have these features land. It sits very well with me; obviously we'd write things differently but I'm happy to have a torrent API in Go that can offer the flexibility a DHT-bot needs.

I may open an issue about the DHT-Sec thing though, because I do think the current approach is on the wrong track.

PR #54, opened 2015-12-29
Mocking  improvements

zhulik:
Those changes allows me to mock network access in my tests. It breaks public API, Client returns Instance of Download interface instead of Torrent. All cmd tools fixed, tests is green


zhulik:
Comments?

anacrolix:
I'll take a look at this shortly.

zhulik:
PR can be merged after your updates, tests still green

.gitignore
```diff
@@ -0,0 +1 @@
+.idea/
```
anacrolix: What's this for?

download.go
```diff
@@ -0,0 +1,27 @@
...
+	SubscribePieceStateChanges() *pubsub.Subscription
+	Seeding() bool
+	SetDisplayName(dn string)
+	Client() *Client
+	AddPeers(pp []Peer) error
+	DownloadAll()
+	Trackers() [][]tracker.Client
+	Files() (ret []File)
+	Peers() map[PeersKey]Peer
+}
```
anacrolix: Missing trailing newline

download.go
```diff
@@ -0,0 +1,27 @@
...
+	BytesCompleted() int64
+	SubscribePieceStateChanges() *pubsub.Subscription
+	Seeding() bool
+	SetDisplayName(dn string)
+	Client() *Client
+	AddPeers(pp []Peer) error
+	DownloadAll()
+	Trackers() [][]tracker.Client
+	Files() (ret []File)
+	Peers() map[PeersKey]Peer
```
anacrolix: Why is Peers() available?

.gitignore
kirillrdy: I think that's intelij

anacrolix:
I think if Download implements everything that Torrent did in an interface fashion, you may be better off just making Torrent into the interface you wanted. Rename the old Torrent type as something private, maybe clientTorrent or something. Are there tests that can be added as a result of these changes? These would make this change worth something, currently it's not particularly useful.

.gitignore
kirillrdy: https://www.jetbrains.com/idea/

zhulik: yes, i use idea with golang plugin. Great tool, try it!

zhulik:
Okey, Download renamed to Torrent, old Torrent renamed to clientTorrent. 

About practical usage: those thanges allows me to mock Client logic and Torrent logic in my code, i can use any realization of Torrent interface in my tests. Replacing Torrent struct with Torrent interface will not force users to change their code, almost.

anacrolix:
I've just started rolling through the changes from merging this through my downstream projects, and it occurred to me, why can't you have an interface that exposes what you need from struct Torrent in your own code rather than putting that interface into the main torrent package? That is to say, your Torrent interface can be in your own mocking code. You can then either use anacrolix/torrent.Torrent, or your own implementation of your Torrent interface.

zhulik:
I have mock interface for Client in my own code
```go
type torrentClient interface {
	Torrents() []torrent.Torrent
	AddMagnet(url string) (torrent.Torrent, error)
	AddTorrent(info *metainfo.MetaInfo) (torrent.Torrent, error)
}
```

So, your client implements it out of the box and i use it 	through this interface . But my mock realization must return real Torrent from Torrents() and Add*() if Torrent is a struct. If Torrent is an interface, i can return my mock realization for Torrent instead of your realization.

anacrolix:
Can't you create an instance of your mock Client that returns your Torrent interface, even for anacrolix/torrent.Client.Torrents? For example:
```
type MockClient interface {
   // all the Client methods you want to use, with the signatures you need for mocking
}

type mockedAnacrolixClient struct {
    *torrent.Client
}

// Clobber the Torrents method for anacrolix/torrent.Client to conform to your mock interface.
func (mockedAnacrolixClient) Torrents() ([]MockTorrent, error) { ... }
```

I appreciate the work in this PR, but I'm struggling to justify it for general use. I can't foresee any situation where anacrolix/torrent would want to return different Torrent implementations, and there are subtle changes in API that have resulted, such as the need to shrink the size of the Torrent interface, as a lot of the methods on the Torrent struct are merely helpers, which aren't appropriate if that type becomes an interface.

zhulik:
As i wrote above, i have client mock with interface
```go
type torrentClient interface {
    Torrents() []torrent.Torrent
    AddMagnet(url string) (torrent.Torrent, error)
    AddTorrent(info *metainfo.MetaInfo) (torrent.Torrent, error)
}
```
If Torrent is a struct, i can't create mock struct that will satisfy this interface and returns fake torrents from `Torrents()`, so i will be forced to use your Torrent realization instead of mock. Trouble here in go's types system, read more [here](http://stackoverflow.com/questions/12994679/golang-slice-of-struct-slice-of-interface-it-implements). In other words, []Torrent is not []TorrentInterface even if Torrent satisfies TorrentInterface.

anacrolix:
I understand it's painful with Go, it's one of my peeves. However I don't see why my solution above won't work. It's important to recognise that you create a MockClient, and MockTorrent interface, and that you still have to wrap the real torrent.Client, and torrent.Torrent to make the necessary tweaks, for example with methods that return []Torrent, as you have pointed out. That method will require a wrapper that converts []Torrent, to []MockTorrent per your examples.

I'm moving the mocking wrapper to separate branch for now.

PR #72, opened 2016-04-03
convenience function for creating magnets

shoenig:


anacrolix:
I like where you're going with this.

PR #81, opened 2016-05-24
fix listenedAddr  bug when TCP or UDP bind dynamic port

lovedboy:
when I disable TCP or UDP and  assign listen address  `x.x.x.x:0`, that means I want to use dynamic port to bind, then `listenedAddr` is not the real address. Anyway I think is a good choice to using `Addr().String()` for get bind real address.


anacrolix:
Really need some tests for this.

anacrolix:
Thanks

PR #82, opened 2016-05-25
add a feature: limit send piece rate

lovedboy:
I want to add a feature which use [token bucket](https://en.wikipedia.org/wiki/Token_bucket)  to limit send piece rate.

### DIFF
--- a/config.go
+++ b/config.go
@@ -40,6 +40,9 @@ type Config struct {
 
 	IPBlocklist iplist.Ranger
 	DisableIPv6 bool `long:"disable-ipv6"`
+	LimitSendPieceRate bool `long:"limit-send-piece-rate"`
+	// how many kB can be send every second
+	SendPieceRate uint32 `long:"max-kbyte-can-send-every-second"`
 	// Perform logging and any other behaviour that will help debug.
 	Debug bool `help:"enable debug logging"`
 }
--- a/client.go
+++ b/client.go
@@ -17,7 +17,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
-
+	"github.com/anacrolix/torrent/ratelimit"
 	"github.com/anacrolix/missinggo"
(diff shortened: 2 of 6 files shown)


config.go
```diff
@@ -40,6 +40,9 @@ type Config struct {
 
 	IPBlocklist iplist.Ranger
 	DisableIPv6 bool `long:"disable-ipv6"`
+	LimitSendPieceRate bool `long:"limit-send-piece-rate"`
+	// how many kB can be send every second
+	SendPieceRate uint32 `long:"max-kbyte-can-send-every-second"`
```
Sioro-Neoku: A simpler approach would be to have just the `SendPieceRate` variable, and have the value `0` to represent no limiting.

Also I'm confused why it's unsigned, negative values have no meaning here.

lovedboy: I will update the pr.

lovedboy:
@anacrolix how about the pr?

anacrolix:
I am considering it.

anacrolix:
Here's a small checklist, I hope you don't mind:

1. Use golang.org/x/time/rate for rate limiting, put this in Client, and copy the pointer in Torrent (so that separate rate limiters can be used for torrents).
3. Make use of the rate limiter directly to decide when to upload chunks. It's no good trying to write them out and due to rate limiting, other control messages are stalled.
4. It's probably best if you let callers just set the rate limiter directly on the client. I don't like Config structs that accumulate ridiculous numbers of switches when the user can just configure those external things separately. Client.SetUploadRateLimiter(...).
5. Create a test like testClientTransfer, that applies rate limiting to the leecher and seeder, and in those circumstances, ensure that the transfer takes an appropriate amount of time.
6. Make it clear that this upload rate limiting is on pieces only. Note that using the copied pointer strategy above, it will make it possible to merge rate limiters as the user desires.
7. Use goimports to format your import blocks.

PR #84, opened 2016-06-01
fix disabling of global bootstrap servers

gdouezangrard:
A missing condition leaves the default root servers to be added even if `NoDefaultBootstrap` is `true` and no custom bootstrap servers are specified.

This pull request adds the missing check.


anacrolix:
Good catch, I should probably redesign some of the wording and methods in this area sometime.

Issue #50, opened 2015-12-23
Improve start-up efficiency when hashing with file-based storage backends


zhulik:
Hi, i built some torrent downloader based on you package. 

When i tryed to download big torrent (about 15GB), i noticed, that memory usage is really great (about 700MB), CPU usage about 80-90% on one core. htop shows me large number of downloader threads (50 at least) and it's increases. 

At about 15% of downloading i see that messages in logs:
`2015/12/23 08:03:00 error writing chunk: open <path to one of files>: too many open files`

My wrapper around your package is small and thin and can't produce described problems. You can find it [here](https://github.com/zhulik/go_torrent_bot)

It is your bugs or it is my fault and i should play with packge config/tuning go runtime?

For example, transmission-web uses about 100MB of memory and 20-30% of one CPU core with 50 seeding torrents and 1-2 active downloads.

I am newbie in go development and not familiar with golang's profilers, but i want to learn it. I will try to debug this problems in the next few days.

Thanks and sorry for my english!

anacrolix:
Thanks for the report. I believe the cause is the way I offload writes to disk which can be too slow for a fast torrent. If you're using a wrapper, could you try the mmap backed data store as an alternative? I'll address the too many open files issue soon. Thanks.

zhulik:
`If you're using a wrapper, could you try the mmap backed data store as an alternative?`

Can you explain?

anacrolix:
In your code at https://github.com/zhulik/go_torrent_bot/blob/master/main.go#L34, you can set TorrentDataOpener to the mmap.TorrentData backend, an example is here in the tests: https://github.com/anacrolix/torrent/blob/master/fs/torrentfs_test.go#L206. This avoids using Open/Close and file handle overhead. Let me know if it's still a problem with this.

zhulik:
Hi, i hade updated torrent package, set mmap.TorrentData as you told me. 
I see, that memory usage still high, htop shows about 500MB, stable.
CPU usage is high too: 100% of first core + 30-50% of second.

Positive changes: i don't see large numbers of threads, just 7, but on 80% of downloading it was increased to 32. Dowloaded files are valid, no too many open files errors.

Negative:
`2015/12/25 08:25:58 Failed to get updates, retrying in 3 seconds...` in logs after downloading

[...]

PR #87, opened 2016-06-17
initial libtorrent

axet:
Hello!

Based on your great work, I decide to create simple go wrapper to make it library friendly. It would be compatible with C, Java, Android (gomobile build) and can have even better future.

It is not finished yet, but may give you some ideas or point me in the right direction. Just want to make my work synchronised with main tree and your preferencies.

I have to change torrent source because your torrent/ sources missing few options:

  * Between application restarts torrent library should not run consistency checks on files, instead it should check one torrent added, and read runtime information from state file. This require additional Check() method accessed by user.
  * Allow to rename files within torrent. it would require keeping to metainfo for network operations and local file storage.
  * Count network bytes for Client and Torrent objects, so UI can show bandwidth information.

And I have few ideas about real time updates for running torrents, current implementations allows pooling only. But for a library would be nice to have realtime updates for changing torrents and not all torrents on the screen. It can be done if library will notify application about every torrent change (for example if chunk downloaded) and application decide if we need to update UI (for example not often then every second) and stay quiet for torrents without events...

Also if main page would have BEP list would be nice....

work in progress...


client.go
```diff
@@ -1531,6 +1531,101 @@ func (cl *Client) AddTorrentInfoHash(infoHash metainfo.Hash) (t *Torrent, new bo
 	return
 }
 
+// Save torrent to state file, metadata + pices states.
+func (cl *Client) SaveTorrent(t *Torrent) []byte {
+	t.cl.mu.Lock()
+	defer t.cl.mu.Unlock()
+
+	return nil
```
anacrolix: I can't accept an incomplete method for now. There is already Torrent.MetaInfo, and the piece state is dependent on the storage backend, and should be relevant there. File/mmap backends for example do require some state to be added, I've indicated my preferred solution for this here https://github.com/anacrolix/torrent/issues/50#issuecomment-167460940.

client.go
```diff
@@ -1531,6 +1531,101 @@ func (cl *Client) AddTorrentInfoHash(infoHash metainfo.Hash) (t *Torrent, new bo
...
+		}
+	}
+	cl.mu.Lock()
+	defer cl.mu.Unlock()
+	if spec.ChunkSize != 0 {
+		t.chunkSize = pp.Integer(spec.ChunkSize)
+	}
+	t.addTrackers(spec.Trackers)
+	return
+}
```
anacrolix: See SaveTorrent.

client.go
```diff
@@ -1531,6 +1531,101 @@ func (cl *Client) AddTorrentInfoHash(infoHash metainfo.Hash) (t *Torrent, new bo
...
+		t.cl.mu.Lock()
+		defer t.cl.mu.Unlock()
+
+		// reopen storage
+		info := t.info
+		t.info = nil
+		t.setInfoBytes(info.Bytes)
+	}
+
+	t.maybeNewConns()
```
anacrolix: I'm not sure what you mean by manually start. Torrent's won't be downloaded until data is requested from them. Excepting Info, which if missing will be fetched from the network.

client.go
```diff
@@ -1531,6 +1531,101 @@ func (cl *Client) AddTorrentInfoHash(infoHash metainfo.Hash) (t *Torrent, new bo
...
+}
+
+// Drop connections from a torrent, prepare to Remove torrent from client or
+// Resume it again. Stop command also cancels check() torrent.
+func (cl *Client) StopTorrent(t *Torrent) {
+	t.close()
+}
```
anacrolix: Just drop the Torrent. I used to have an intermediate "cease networking" state, and it's not worth it.

client.go
```diff
@@ -1531,6 +1531,101 @@ func (cl *Client) AddTorrentInfoHash(infoHash metainfo.Hash) (t *Torrent, new bo
...
+		t.info = nil
+		t.setInfoBytes(info.Bytes)
+	}
+
+	t.check(func() {
+		if restart {
+			cl.StartTorrent(t)
+		}
+	})
+}
```
anacrolix: This may have some value. If user code knows that the data state may be incorrect or have changed considerably, we may want to force a recheck without opening a reader and scanning through the whole torrent ourselves.

axet:
I want to publish library under LGPL3.0. I will create separate repo and keep only changes to your library here

PR #89, opened 2016-07-01
Few features

axet:
Based on your great work, I decide to create simple go wrapper to make it library friendly. It would be compatible with C, Java, Android (gomobile build) and can have even better future.

It is not finished yet, but may give you some ideas or point me in the right direction. Just want to make my work synchronised with main tree and your preferencies.

I have to change torrent source because your torrent/ sources missing few options:

- Between application restarts torrent library should not run consistency checks on files, instead it should check one torrent added, and read runtime information from state file. This require additional Check() method accessed by user.

- Count network bytes for Client and Torrent objects, so UI can show bandwidth information.

- Better network management requires to control when torrent active or paused. And few methods for configure port's for UPNP.

You can see you library (with above changes) in action here:

https://play.google.com/store/apps/details?id=com.github.axet.torrentclient

EDIT: current cmd/* should be updated to support new client.Start()

### DIFF
--- a/client.go
+++ b/client.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"crypto/rand"
 	"encoding/hex"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -62,7 +63,9 @@ type Client struct {
 	halfOpenLimit int
 	peerID        [20]byte
 	// The net.Addr.String part that should be common to all active listeners.
-	listenAddr     string
+	listenAddr string
+	// listen UDP addr, may be different from TCP port because of UPNP
+	listenUDPAddr  string
 	tcpListener    net.Listener
 	utpSock        *utp.Socket
 	dHT            *dht.Server
@@ -82,6 +85,11 @@ type Client struct {
 	closed missinggo.Event
 
 	torrents map[metainfo.Hash]*Torrent
+
+	// Since we have to have loadTorrent, we woudl like to keep other statistics like all modern torrent apps does.
+	// Usint this data we will be able to measure Client total bandwidth.
+	downloaded int64
+	uploaded   int64
 }
 
 func (cl *Client) IPBlockList() iplist.Ranger {
@@ -303,14 +311,20 @@ func NewClient(cfg *Config) (cl *Client, err error) {
 	if err != nil {
 		return
 	}
+
+	return
+}
+
+// for UPNP handling, do not start DHT before we know ours external ip/port
+func (cl *Client) Start() (err error) {
 	if cl.tcpListener != nil {
 		go cl.acceptConnections(cl.tcpListener, false)
 	}
 	if cl.utpSock != nil {
 		go cl.acceptConnections(cl.utpSock, true)
 	}
-	if !cfg.NoDHT {
-		dhtCfg := cfg.DHTConfig
+	if !cl.config.NoDHT {
+		dhtCfg := cl.config.DHTConfig
 		if dhtCfg.IPBlocklist == nil {
 			dhtCfg.IPBlocklist = cl.ipBlockList
 		}
@@ -323,7 +337,6 @@ func NewClient(cfg *Config) (cl *Client, err error) {
 			return
 		}
 	}
-
 	return
 }
 
@@ -563,12 +576,17 @@ func (cl *Client) dialFirst(addr string, t *Torrent) (conn net.Conn, utp bool) {
 	return
 }
 
-func (cl *Client) noLongerHalfOpen(t *Torrent, addr string) {
+func (cl *Client) noLongerHalfOpen(t *Torrent, addr string) bool {
+	if t.halfOpen == nil {
+		return false
+	}
 	if _, ok := t.halfOpen[addr]; !ok {
-		panic("invariant broken")
+		// halfOpen been closed, drop all past connections
+		return false
 	}
 	delete(t.halfOpen, addr)
 	cl.openNewConns(t)
+	return true
 }
 
 // Performs initiator handshakes and returns a connection. Returns nil
@@ -633,7 +651,13 @@ func (cl *Client) outgoingConnection(t *Torrent, addr string, ps peerSource) {
 	defer cl.mu.Unlock()
 	// Don't release lock between here and addConnection, unless it's for
 	// failure.
-	cl.noLongerHalfOpen(t, addr)
+	if !cl.noLongerHalfOpen(t, addr) {
+		// connection come after torrent was restarted, drop connection
+		if c != nil {
+			c.Close()
+		}
+		return
+	}
 	if err != nil {
 		if cl.config.Debug {
 			log.Printf("error establishing outgoing connection: %s", err)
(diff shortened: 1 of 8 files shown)


anacrolix:
Thanks for submitting this back. It's a lot to digest and I'll do it as I can.

client.go
```diff
@@ -62,7 +63,9 @@ type Client struct {
 	halfOpenLimit int
 	peerID        [20]byte
 	// The net.Addr.String part that should be common to all active listeners.
-	listenAddr     string
+	listenAddr string
+	// listen UDP addr, may be different from TCP port because of UPNP
+	listenUDPAddr  string
```
anacrolix: Peers expect the TCP and UDP ports to be the same. How do you announce to DHT, trackers, and over PEX with this?

client.go
```diff
@@ -1524,12 +1576,235 @@ func (cl *Client) AddTorrentInfoHash(infoHash metainfo.Hash) (t *Torrent, new bo
...
+	// time
+	DownloadingTime int64 `json:"downloading_time,omitempty"`
+	SeedingTime     int64 `json:"seeding_time,omitempty"`
+
+	// .torrent
+	Comment   string `json:"comment,omitempty"`
+	Creator   string `json:"creator,omitempty"`
+	CreatedOn int64  `json:"created_on,omitempty"`
+}
+
```
anacrolix: All this state stuff seems specific to your application, and can be done external to this package.

client.go
```diff
@@ -1524,12 +1576,235 @@ func (cl *Client) AddTorrentInfoHash(infoHash metainfo.Hash) (t *Torrent, new bo
...
+	t.addedDate = s.AddedDate
+	t.completedDate = s.CompletedDate
+
+	t.comment = s.Comment
+	t.creator = s.Creator
+	t.createdOn = s.CreatedOn
+
+	return
+}
+
```
anacrolix: See last.

client.go
```diff
@@ -1524,12 +1576,235 @@ func (cl *Client) AddTorrentInfoHash(infoHash metainfo.Hash) (t *Torrent, new bo
...
+// run file consictency checks and then manually start it. Here are two
+// possibilites: 1) Create *Torrent without adding it to client. 2) Add Torrent
+// to client and do not announce it or download.
+func (cl *Client) StartTorrent(t *Torrent) error {
+	cl.mu.Lock()
+	defer cl.mu.Unlock()
+
+	return cl.startTorrent(t)
+}
+
```
anacrolix: No downloading occurs other than to fetch metainfo until the Torrent is instructed which pieces are wanted. It seems that you want network activity, even just for the metainfo, even in your case?

client.go
```diff
@@ -1872,3 +2178,29 @@ func (cl *Client) banPeerIP(ip net.IP) {
...
 	cl.badPeerIPs[ip.String()] = struct{}{}
 }
+
+func (cl *Client) Stats() (downloaded int64, uploaded int64) {
+	cl.mu.Lock()
+	defer cl.mu.Unlock()
+	downloaded = cl.downloaded
+	uploaded = cl.uploaded
+	return
+}
```
anacrolix: This would be better as a block. Return a ClientStats struct. There will be more values.

connection.go
```diff
@@ -76,6 +76,8 @@ type connection struct {
 	// The peer has everything. This can occur due to a special message, when
 	// we may not even know the number of pieces in the torrent yet.
 	peerHasAll bool
+	// number pieces completed
+	peerCompleted int
```
anacrolix: I think this value can be inferred from the peerHas Bitmap or peerHasAll. It should be a method.

t.go
```diff
@@ -203,3 +219,263 @@ func (t *Torrent) AddTrackers(announceList [][]string) {
...
+	// byte info information
+	Downloaded int64
+	Uploaded   int64
+}
+
+func (t *Torrent) PeersCount() int {
+	t.cl.mu.Lock()
+	defer t.cl.mu.Unlock()
+	return len(t.peers)
+}
```
anacrolix: Stats block.

t.go
```diff
@@ -203,3 +219,263 @@ func (t *Torrent) AddTrackers(announceList [][]string) {
...
+			}
+		}
+		if len(n) > 0 {
+			nn = append(nn, n)
+		}
+	}
+
+	t.metainfo.AnnounceList = nn
+}
+
```
anacrolix: This is needed, good.

t.go
```diff
@@ -203,3 +219,263 @@ func (t *Torrent) AddTrackers(announceList [][]string) {
...
+	}
+	return v
+}
+
+func (t *Torrent) PiecePended(i int) bool {
+	t.cl.mu.Lock()
+	defer t.cl.mu.Unlock()
+
+	return t.pendingPieces.Contains(i)
+}
```
anacrolix: I think there's already a piece state method that should be augmented with this, rather than adding a separate function.

t.go
```diff
@@ -203,3 +219,263 @@ func (t *Torrent) AddTrackers(announceList [][]string) {
...
+		now := time.Now().Unix()
+		if t.seeding() {
+			seeding = seeding + (now - t.activateDate)
+		} else {
+			downloading = downloading + (now - t.activateDate)
+		}
+	}
+	return
+}
+
```
anacrolix: All the above methods belong in one method with a stats struct.

t.go
```diff
@@ -203,3 +219,263 @@ func (t *Torrent) AddTrackers(announceList [][]string) {
...
+func (t *Torrent) Dates() (added int64, completed int64) {
+	added = t.addedDate
```
anacrolix: This isn't the concern of this package.

torrent.go
```diff
@@ -248,21 +300,73 @@ func (t *Torrent) setInfoBytes(b []byte) error {
...
 			t.verifyPiece(i)
+			if t.closed.IsSet() {
+				return
+			}
 		}
 	}()
+}
+
+// for big torrent we need to save CPU time and do not recheck files on disk
+// bettwen application restarts. loadTorrent allows to restore pieces state
```
anacrolix: I mentioned this before, this should be done by augmenting the appropriate storage backend. I recently added this to the file and mmap backends, you could test they have the desired effect on your platform.

anacrolix:
The tracker changes make a lot of sense. We need to clarify the start/stopping of Torrents. The extra stats should be added, but using structs the way I mentioned so as not to create a method for every single one. We also need to clarify the UDP/TCP port difference, and how to do announces due to that. Managing the PMP/NAT traversal from within this package would actually be appropriate I believe, as the package also manages those same ports it wants to tunnel. Some of the state stuff can be moved out.

Are you able to make each of these changes as smaller PRs? If not, I'll probably merge then evict the parts that don't fit, and clean up. Thanks.

axet:
Just take what you feel is right and I will adopt the remaining.

You have to be in charge of your library, and reject all you do not like or do not understand. Otherwise the library will be ruined.

I would like to see more basic stats supported by your code, if not, I keep it in my library. This basic information about torrent statistics and source torrent file, one of it currently lost and then regenerated with not relative information.

I do not really understand your point right now, but you can see my by analysing source code. I'm trying to make your code more features which is expected from torrent library. And pushing more control over network activity (including control for metadata connections - yes I think this is important), torrent content and statistics.

I did not know about UDP/TCP tracker limitation (and it seems like bad design), keep it away for a while.

Lets see what you want to take and how you wan't to implement it and I'll update the rest.

anacrolix:
Cheers, will do. I'll merge what I can over the next few days.

axet:
I see what you telling me. I will move most counters and states to my library with your mechanism of TorrentOpener mechanics. And keep tracker changes and will make a few more public methods here.

t.go
axet: @anacrolix not sure what method are you talking about

PR #90, opened 2016-07-04
Less features

axet:
- Between application restarts torrent library should not run consistency checks on files, instead it should check one torrent added, and read runtime information from state file. This require additional Check() method accessed by user.

- Count network bytes for Client and Torrent objects, so UI can show bandwidth information.

- Better network management requires to control when torrent active or paused. And few methods for configure port's for UPNP.

### DIFF
--- a/global.go
+++ b/global.go
@@ -6,6 +6,10 @@ import (
 	"time"
 )
 
+var (
+	ExtendedHandshakeClientVersion = extendedHandshakeClientVersion
+)
+
 const (
 	pieceHash        = crypto.SHA1
 	maxRequests      = 250    // Maximum pending requests we allow peers to send us.
--- a/t.go
+++ b/t.go
@@ -4,9 +4,11 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/anacrolix/missinggo/bitmap"
 	"github.com/anacrolix/missinggo/pubsub"
 
 	"github.com/anacrolix/torrent/metainfo"
+	pp "github.com/anacrolix/torrent/peer_protocol"
 )
 
 // The torrent's infohash. This is fixed and cannot change. It uniquely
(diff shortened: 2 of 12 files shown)


torrent.go
```diff
@@ -37,6 +38,12 @@ type peersKey struct {
 	Port    int
 }
 
+type TorrentStats struct {
+	Downloaded int64
+	Uploaded   int64
+	PeersCount int
+}
```
anacrolix: Put the PEX and DHT stats here.

torrent.go
```diff
@@ -73,8 +80,18 @@ type Torrent struct {
...
 	// An announcer for each tracker URL.
 	trackerAnnouncers map[string]*trackerScraper
+
 	// How many times we've initiated a DHT announce.
 	numDHTAnnounces int
+	// how many peers we got, not sure if we have to use last value or increment
+	peersDHT int
+	// error message
+	errDHT string
+	// last call
```
anacrolix: Move these to the torrent stats

t.go
```diff
@@ -203,3 +218,157 @@ func (t *Torrent) AddTrackers(announceList [][]string) {
...
+				T.Peers = trackerScraper.peers
+				T.Err = trackerScraper.err
+				T.LastAnnounce = trackerScraper.lastAnnounce
+				T.NextAnnounce = trackerScraper.nextAnnounce
+			}
+			tt = append(tt, T)
+		}
+	}
+
+	tt = append(tt, Tracker{"DHT", t.peersDHT, t.errDHT, t.lastAnnounceDHT, 0})
```
anacrolix: These should be gotten from Torrent.Stats, rather than forcing them to fit into Tracker stats

global.go
```diff
@@ -36,7 +36,7 @@ const (
 	// http://www.bittorrent.org/beps/bep_0005.html
 	defaultExtensionBytes = "\x00\x00\x00\x00\x00\x10\x00\x01"
 
-	socketsPerTorrent     = 80
+	socketsPerTorrent     = 40
```
anacrolix: Make this configurable instead. Torrent.SetMaxConns.

t.go
axet: This is heavy operation. Do we need fill trackers into stat struct even if it not needed?

metainfo/metainfo.go
```diff
@@ -8,12 +8,24 @@ import (
...
 
 	"github.com/anacrolix/torrent/bencode"
 )
 
+type FilesSlice []FileInfo
+
+func (p FilesSlice) Len() int { return len(p) }
+func (p FilesSlice) Less(i, j int) bool {
+	pi := strings.Join(p[i].Path, string(filepath.Separator))
+	pj := strings.Join(p[j].Path, string(filepath.Separator))
```
anacrolix: Do you know if there's a specified sort order in any specification?

axet: I found sort algorithm in transmissionbt. And to make same files produce identical metainfo I add sort.

https://github.com/wereHamster/transmission/blob/master/libtransmission/makemeta.c

qsort -> builderFileCompare -> asc

anacrolix:
I'm nearly done making tracker status available in anacrolix:master. Do trackerScraper and announceRequestResult contain all the infos you'd need?

tracker_scraper.go
```diff
@@ -69,6 +73,7 @@ func (me *trackerScraper) announce() (ret trackerAnnounceResult) {
 		ret.Completed = time.Now()
 	}()
 	ret.Interval = 5 * time.Minute
+	ret.Err = nil
```
axet: @anacrolix check this line. important to clear error.

axet:
@anacrolix yes. I already rebased it. thanks!

axet:
Also I want to attract your attention to closed torrents. Regardless If they were closed (.Drop()) they will become active when .addPeer() will be called. Which can be called from may places halfopen connections, trackers, maybe more places ...

Easy to debug and of course easiest way to do so drop t.peers = nil and see what will happens.

I'm saying it because seems like you do not want merge this pull request which includes additional control over network torrents/client state.

anacrolix:
The PR is/was very large, and its difficult to reconcile the impact that lots of unrelated changes will have. I see you've recently removed the active torrent stuff, which is positive in my mind, I'd like to accept this PR if the start/stop stuff isn't included. If it's easier, perhaps make another PR containing the changes without the start/stop stuff (but definitely including the fixes for closed/dropped torrents), I expect we can get it merged with only a few small changes or clarifications.

PR #98, opened 2016-07-19
close cleanups fixes. care more about .Drop() torrents, do not start …

axet:
…it again when half open connection bring new addPeers()

### DIFF
--- a/client.go
+++ b/client.go
@@ -310,8 +310,11 @@ func NewClient(cfg *Config) (cl *Client, err error) {
 	if cl.utpSock != nil {
 		go cl.acceptConnections(cl.utpSock, true)
 	}
-	if !cfg.NoDHT {
-		dhtCfg := cfg.DHTConfig
+	if !cl.config.NoDHT {
+		if cl.dHT != nil {
+			return
+		}
+		dhtCfg := cl.config.DHTConfig
 		if dhtCfg.IPBlocklist == nil {
 			dhtCfg.IPBlocklist = cl.ipBlockList
 		}
@@ -345,12 +348,15 @@ func (cl *Client) Close() {
 	cl.closed.Set()
 	if cl.dHT != nil {
 		cl.dHT.Close()
+		cl.dHT = nil
 	}
 	if cl.utpSock != nil {
 		cl.utpSock.Close()
+		cl.utpSock = nil
 	}
 	if cl.tcpListener != nil {
 		cl.tcpListener.Close()
+		cl.tcpListener = nil
 	}
 	for _, t := range cl.torrents {
 		t.close()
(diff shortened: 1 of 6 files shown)


client.go
```diff
@@ -310,8 +310,11 @@ func NewClient(cfg *Config) (cl *Client, err error) {
...
 		go cl.acceptConnections(cl.utpSock, true)
 	}
-	if !cfg.NoDHT {
-		dhtCfg := cfg.DHTConfig
+	if !cl.config.NoDHT {
+		if cl.dHT != nil {
+			return
+		}
+		dhtCfg := cl.config.DHTConfig
 		if dhtCfg.IPBlocklist == nil {
```
anacrolix: I don't think we need this.

client.go
```diff
@@ -345,12 +348,15 @@ func (cl *Client) Close() {
...
 	}
 	if cl.utpSock != nil {
 		cl.utpSock.Close()
+		cl.utpSock = nil
 	}
 	if cl.tcpListener != nil {
 		cl.tcpListener.Close()
+		cl.tcpListener = nil
 	}
 	for _, t := range cl.torrents {
```
anacrolix: None of the above have any effect.

client.go
```diff
@@ -564,12 +570,18 @@ func (cl *Client) dialFirst(addr string, t *Torrent) (conn net.Conn, utp bool) {
 	return
 }
 
-func (cl *Client) noLongerHalfOpen(t *Torrent, addr string) {
+func (cl *Client) noLongerHalfOpen(t *Torrent, addr string) bool {
```
anacrolix: I think I'd rather just leave the halfOpen field even when the torrent is closed. No need to return anything. Just check for torrent closed when it matters?

torrent.go
```diff
@@ -1252,6 +1263,11 @@ func (t *Torrent) announceDHT(impliedPort bool) {
 					allAddrs[key] = struct{}{}
 				}
 				cl.mu.Lock()
+				if t.peers == nil { // torrent can be already closed. return.
```
anacrolix: I don't think you should use zeroed state to indicate something being closed. Just test Torrent.closed directly if it matters. Here, Torrent.addPeers should just not trigger methods like openNewConns etc. if it's closed.

torrent.go
```diff
@@ -532,11 +532,17 @@ func (t *Torrent) close() (err error) {
...
+	t.trackerAnnouncers = nil
 	for _, conn := range t.conns {
 		conn.Close()
 	}
+	t.conns = nil
+	t.peers = nil
+	t.halfOpen = nil
 	t.pieceStateChanges.Close()
-	t.updateWantPeersEvent()
 	return
```
anacrolix: Rather than the above changes, trackerScrapers should just test for Torrent.closed in their main loop. Torrent.addPeers should not trigger new network activity if the Torrent is closed. Torrent.updateWantPeersEvent should still occur, but wantPeers is always false if the torrent is closed.

tracker_scraper.go
```diff
@@ -74,6 +74,7 @@ func (me *trackerScraper) announce() (ret trackerAnnounceResult) {
 		ret.Completed = time.Now()
 	}()
 	ret.Interval = 5 * time.Minute
+	ret.Err = nil
```
anacrolix: This is redundant, ret.Err is already nil.

client.go
axet: it does when code start to live its own live. and when it happens it will crash. it self checking nil. for example when you .Drop() torrent it will start live it's own live after receiving half open connection data or tracker announces.

client.go
axet: same as above. self checking code. when you start to control torrent behaviour after you drop it, this code matters - torrent will behave property when something happens after you close it.

torrent.go
axet: True. It additional runtime checks of proper behaviour of closed torrents. Should we call addPeer on closed torrent? I think no.

tracker_scraper.go
axet: Lets pretend we want to announce second time. Still want to keep pervious error?

axet:
Seems like we see it differently. It is fine. I prefer nil, sine it will show at runtime with panic if I failed understanding live cycle of object.

PR #105, opened 2016-08-01
fs: fix isSubPath for top-level directories

### LINKED ISSUE #104: torrentfs: directories show up empty in torrents with dirs
If you download any torrent that contains directories (e.g.: https://archive.org/download/TheFastandtheFuriousJohnIreland1954goofyrip/TheFastandtheFuriousJohnIreland1954goofyrip_archive.torrent) the top-level directories show up but nothing bellow them.

perses:
I return false if parent="" and child="" only to satisfy the test (not sure when this happens). Otherwise I would have made it `if len(parent) == 0 { return true }`

Fixes #104

### DIFF
--- a/fs/torrentfs.go
+++ b/fs/torrentfs.go
@@ -158,6 +158,9 @@ var (
 )
 
 func isSubPath(parent, child string) bool {
+	if len(parent) == 0 {
+		return len(child) > 0
+	}
 	if !strings.HasPrefix(child, parent) {
 		return false
 	}
--- a/fs/torrentfs_test.go
+++ b/fs/torrentfs_test.go
@@ -214,6 +214,7 @@ func TestIsSubPath(t *testing.T) {
 	}{
 		{"", "", false},
 		{"", "/", true},
+		{"", "a", true},
 		{"a/b", "a/bc", false},
 		{"a/b", "a/b", false},
 		{"a/b", "a/b/c", true},


anacrolix:
Did you have a case that triggered the need for this change?

perses:
Yes, I mentioned a torrent in the bug. Basically all torrents that I've tried that contain a directory. All top-level entries show up but nothing in the directories.

perses:
Now that I think about it...

In `func (dn dirNode) ReadDirAll` I was getting dn.path="" and the files were "TheFastandtheFuriousJohnIreland1954goofyrip_512kb.mp4", etc, even though they were all inside the dir "TheFastandtheFuriousJohnIreland1954goofyrip".

Shouldn't dn.path have been "TheFastandtheFuriousJohnIreland1954goofyrip" and the files "TheFastandtheFuriousJohnIreland1954goofyrip/TheFastandtheFuriousJohnIreland1954goofyrip_512kb.mp4", etc?

perses:
Either way, not isSubPath behaves the same way as before e735aeeee381426631e3fd8b01a30203236e9417.

anacrolix:
I don't remember, it's been a really long time :D. Feel free to add any tests that ensure it's correct for your case.

perses:
Right now it seems that everything works as it should. I added a testcase for this in torrentfs_test.go. If I find anything else I'll let you know.

PR #120, opened 2016-09-21
fix dht-server cmd won't exit after Ctrl+C

mengzhuo:
current dht-server will stay running even SIGTERM sended


anacrolix:
I cherry picked your commits c4979954e86cf73f53426d81902bb52689c20862 and 23efae91e2620886659c05f9e213f69c0e257392, you had some bad rebasing going on.

PR #122, opened 2016-10-08
Add download/upload rate limit

### LINKED ISSUE #121: Throttle download/upload speed
Preventing torrent consumed all network bandwidth,  download/upload speed throttle options are needed.

hustcat:
Add download/upload rate limit. fix #121

### DIFF
--- a/client.go
+++ b/client.go
@@ -31,6 +31,7 @@ import (
 	"github.com/anacrolix/torrent/mse"
 	pp "github.com/anacrolix/torrent/peer_protocol"
 	"github.com/anacrolix/torrent/storage"
+	"golang.org/x/time/rate"
 )
 
 // Currently doesn't really queue, but should in the future.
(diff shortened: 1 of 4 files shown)


axet:
I do not think using time.Sleep is a good idea. You better to use socket window feature. For example do not read from socket for specified period of time (not sleeping, but not reading).

But let see what anacrolix thinks...

anacrolix:
It's a good attempt. I don't think you want to sleep while the client mutex is held. You need to consider the cases where a peer disconnects before we upload to them. That upload to some peers will starve others. That cancelled requests, that are removed from the connection write queue are also canceled from the limiter. If you can have a limiter with an infinite rate, that's preferable to introducing a nil case that has to be checked everywhere too.

hustcat:
@anacrolix  Sleep in downloadChunk is not a good idea. I made some changes.

anacrolix:
@hustcat 

I added a very primitive but functionally complete (I think) upload rate limiter. https://github.com/anacrolix/torrent/commit/836bb344f3602205275f9f6abbe9a986bb988e41

PR #126, opened 2016-10-19
Trivial HTTP support for adding torrent files by

teran:
Sometimes it's really usefull to do not download
torrent file locally but simply pass HTTP link
this patch adds such ability.

Signed-off-by: Igor Shishkin <me@teran.ru>


anacrolix:
Seems fair, thanks.

PR #128, opened 2016-11-04
feat: remove print BuildFromFilePath

Shamanoid:
removes the print of (path, err) in function BuildFromFilePath


Shamanoid:
It can become super noisy if you have dozens of files in the directory :)

PR #140, opened 2016-12-29
Add CLI flag to abort verify at first sign of trouble.

BonkaBonka:
I wanted torrent-verify to bail out when it finds an issue rather than continuing on and checking everything.  The panic also has a nice side-effect of setting a non-zero return code so that a wrapper script can trap for it.


anacrolix:
I've made this the default. Thanks.

BonkaBonka:
Thanks for fixing #141!  During the rewrite though, I think you might've forgotten to do the fast-fail in the case of a piece mismatch.  :)  I can generate a pull request against the new code if you'd like.

anacrolix:
Please do

PR #144, opened 2017-01-04
Abort verify at first hash mismatch.

BonkaBonka:
Per #140, I've updated the patch (such as it is) to work with the rewrite.


BonkaBonka:
Perfection!  Thank you so much!

PR #149, opened 2017-03-16
fix `Spec.Storage` and allow per-torrent dir

thechriswalker:
`TorrentSpec.Storage` was not honored when calling `Client.AddTorrentSpec`
and the configured `cfg.DefaultStorage` was always used. Now if you construct
your `TorrentSpec` you can specify any `StorageImpl`

Also, the most common use case for custom storage being per-torrent paths for
FileStorage, this adds a `pathMaker` function to the File implementation that
allows customization, along with the default (always use base path) and my use
case (which seemed common enough from the Gitter chat) which is infohash based
subdirectories.

All Public methods have not changed signature, but 1 private method did, hence
the test update.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1124,7 +1124,13 @@ func (cl *Client) badPeerIPPort(ip net.IP, port int) bool {
 }
 
 // Return a Torrent ready for insertion into a Client.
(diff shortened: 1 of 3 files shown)


anacrolix:
Great work.

PR #153, opened 2017-04-20
Add public methods for torrent peer counts

joelanford:
I'd need access to peer counts, so I added 4 new methods on the Torrent struct to provide them.

`NumActivePeers`
`NumHalfOpenPeers`
`NumPendingPeers`
`NumTotalPeers`

The first three are straightforward and return the length of the underlying map (`t.conns`, `t.halfOpen`, and `t.peers`, respectively)

`NumTotalPeers` builds a map containing the set of all peer addresses and returns the length of that map, thus ensuring peers present in multiple maps aren't double counted.

### DIFF
--- a/torrent.go
+++ b/torrent.go
@@ -1271,7 +1271,28 @@ func (t *Torrent) addPeers(peers []Peer) {
 func (t *Torrent) Stats() TorrentStats {
 	t.cl.mu.Lock()
 	defer t.cl.mu.Unlock()
+
+	t.stats.ActivePeers = len(t.conns)
+	t.stats.HalfOpenPeers = len(t.halfOpen)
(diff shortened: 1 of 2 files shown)


anacrolix:
Do you think this could be added to Torrent.Stats instead? The Torrent method count is already very high.

joelanford:
I suppose it could. Are you suggesting that new fields for the counts be added to the TorrentStats struct (but not ConnStats) and then have the counts updated at the same time the Torrent peer maps are updated?

anacrolix:
No I'd just populate the dynamic TorrentStats values when Torrent.Stats is returned.

joelanford:
Pushed new updates, which moves peer stats to the TorrentStats struct.

PR #154, opened 2017-05-10
bug fix: race condition between outgoing connection and incoming conn…

mlmhl:
I think there is a race condition between `incomingConnection` and `outgoingConnection`, as `outgoingConnection` creates the connection(in `establishOutgoingConn`) first and then invokes `runHandshookConn`, which will close the connection if the peer has received a connection from the same peer. Since these two operation(`establishOutgoingConn` and `runHandshookConn`) are not atomic, maybe both peer will close the connetion.
Think about the example as follows: Peer A find Peer B from tracker and Peer B found Peer A too. Both peer invoke `establishOutgoingConn`, which opens a connection to the other one, and each peer invokes `incomingConnection` and record the connection. then each peer invokes `runHandshookConn` after the `establishOutgoingConn` operation, and both find the connection opened by the other side, in this condition, both peers will close the connetion opened by itself, and causes no available connection between the two peers.

solution: In `runHandshookConn`, if we found a connection already exist with the same peers, close the conenction from peer with low ID to the higher one. This promise both peers close the same connection, and retain the other one.

### DIFF
--- a/client.go
+++ b/client.go
@@ -899,7 +899,7 @@ func (cl *Client) runInitiatedHandshookConn(c *connection, t *Torrent) {
 		cl.dopplegangerAddrs[addr] = struct{}{}
 		return
 	}
-	cl.runHandshookConn(c, t)
+	cl.runHandshookConn(c, t, true)
 }
 
 func (cl *Client) runReceivedConn(c *connection) {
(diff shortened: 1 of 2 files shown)


anacrolix:
Do you know of anywhere that mentions this circumstance in the specs?

Perhaps you could add a counter for dropping an existing connection due to the peer ID ordering you introduce?

Can you foresee any situations where this behaviour might be abused? What if I set my peer ID low or high to intentionally be the receiver or initiator in these circumstances?

mlmhl:
hi, @anacrolix thanks for your question.

Here I just use Peer ID to help two peers to reach agreement on which connection to use if they all open a connection to the other one simultaneously.

For example: Peer A open a new connection `1` to Peer B, and at the same time, Peer B open a new to Peer A, In the current implementation, both peers maybe close the connection opened by itself(as mensioned earlier), result in losting contact between Peer A and Peer B. I think this is not a reasonable operation, so we need a machanism to determine using which connection, and then close the other one. As peer's ID is immutable and unique, so I use peer ID to identify different connections. For the example as above, connection `1`'s creater is Peer A, receiver is Peer B, and connection `2`'s creater is Peer B, and receiver is Peer A. If Peer A's ID is lower than Peer B's, Peer B will close connection `2`, and vice versa. It can promise only one connection is retained.

anacrolix:
Looks good, I'll do some testing with it before I merge.

PR #156, opened 2017-05-19
make completion configurable

mlmhl:
In my environment, store the pieces completion information in memory is enough, so I need a method to create file storage and mmap storage with compeltion type.

### DIFF
--- a/storage/boltPieceCompletion.go
+++ b/storage/boltPieceCompletion.go
@@ -18,7 +18,7 @@ type boltPieceCompletion struct {
 	db *bolt.DB
 }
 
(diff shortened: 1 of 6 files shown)


anacrolix:
Uff, why not make pieceCompletion a parameter? Then you can pass whatever you want.

mlmhl:
Thanks @anacrolix , this is a good idea, I will update the commit and submit it later.

mlmhl:
hi @anacrolix I have make pieceCompletion as a parameter, maybe you want to reveiw this.

anacrolix:
Great work.

PR #161, opened 2017-06-05
signal an event when torrent closed

mlmhl:
We need to broadcast an event when a torrent is closed, as other goroutines(such as the torrent `Reader`) maybe waiting for this.

### DIFF
--- a/torrent.go
+++ b/torrent.go
@@ -553,6 +553,7 @@ func (t *Torrent) close() (err error) {
 	for conn := range t.conns {
 		conn.Close()
 	}
+	t.cl.event.Broadcast()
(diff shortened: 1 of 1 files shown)


anacrolix:
Are you seeing any routines wait for this in particular? I think I was trying to move away from the Client.event variable, as it's a bit heavy when everything synchronises on the one value.

mlmhl:
Hi @anacrolix , In my practice, I use a `Reader` of the torrent, and if the torrent is closed, the `Read` method of this `Reader` will blocked as the `waitReadable` function is waiting for the Client.event variable.

PR #164, opened 2017-06-21
guarantee randomness if reading data while downloading

mlmhl:
In my practice, I use a Reader to read torrent data sequentially while the torrent is downloading. In this condition, we reduce the priority of the piece which is reading right now, hope this piece will be download as soon as possible. However, this will result in sequential downloads as I read the data sequentially, loose randomness  between different connections. What about make the piece is reading right now has same priority with the pieces will be read ahead soon. This will promise some randomness  and the piece also can be available as soon as possible.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -549,7 +549,7 @@ func (cn *connection) updatePiecePriority(piece int) {
 	case PiecePriorityReadahead:
 		prio -= cn.t.numPieces()
 	case PiecePriorityNext, PiecePriorityNow:
-		prio -= 2 * cn.t.numPieces()
+		prio -= cn.t.numPieces()
 	default:
 		panic(tpp)
 	}


anacrolix:
That's interesting that you've decided to increase the randomness, while other people like @dimitriss are trying to go the other way. If you consume a Reader as quickly as the data arrives, it's sequential anyway (unless some connections are missing the "now" and "next" pieces).

I have been toying with the idea of making the current best connection, always sequential, (i.e. Nows, then Nexts, then Readaheads, with no deviation from that ordering, and always lower index pieces first) and making the rest act more like you've made the default in this PR. I believe that would satisfy all parties more effectively.

mlmhl:
@anacrolix  it sounds like a pretty good solution, my usage scenarios are special and not very universal, looking forward to the birth of your new program! Now I will close this PR.

anacrolix:
Sure, thanks for taking the dive.

Issue #183, opened 2017-09-01
Implement `Peers()` for type Torrent


boramalper:
I need to move a torrent between different `Client`s in my program and to do this, being able to copy the list of the active peers is sufficient. But unfortunately, there is no `Peers()` function for type Torrent that returns the list of all `Peer`s for a given torrent.

I can implement it myself and send a PR if you are fine.

Best!

anacrolix:
Sounds good. I'm not sure of the naming to use, as you have "pending", "active", and "half-open" connections to peers, but sounds like you only intend to return `Peer` values for them. That type can be produced for each of those peer categories from the data structures in `Torrent`, and the combined state for them represents a subset of the swarm for the torrent, the known swarm if you will. With that information I'd suggest adding `func (*Torrent) KnownSwarm() []Peer`, returning `Peer` values derived from all the aforementioned states.

christian-roggia:
+1 I was looking for a function similar to this one in the past

anacrolix:
@boramalper Have you started on this? Do you want me to do it?

boramalper:
@anacrolix I'm starting right now =)

anacrolix:
I believe this is fixed!

PR #188, opened 2017-09-12
added the initial version of the (t *Torrent) KnownSwarm() function

boramalper:
For #183. Please check `TODO`s before/after merging.

### DIFF
--- a/torrent.go
+++ b/torrent.go
@@ -131,6 +131,11 @@ func (t *Torrent) KnownSwarm() (ks []Peer) {
 		ks = append(ks, peer)
 	}
 
+	// Add half-open peers to the list
+	for _, peer := range t.halfOpen {
+		ks = append(ks, peer)
+	}
+
 	// Add active peers to the list
 	for conn := range t.conns {
 		host, portString, err := net.SplitHostPort(conn.remoteAddr().String())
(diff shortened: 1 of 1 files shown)


torrent.go
```diff
@@ -122,6 +123,44 @@ func (t *Torrent) Closed() <-chan struct{} {
...
+			Port: port,
+			Source: conn.Discovery,
+			// TODO: the connection can be unencrypted due to our (or the peer's) preference,
+			//       but the remote peer might support the encryption. Find a better way to query
+			//       that information, if possible.
+			SupportsEncryption: conn.encrypted,
+		})
+	}
+
+	// TODO: how can we add half-open peers?
```
anacrolix: You can get the addresses from Torrent.halfOpen.

torrent.go
```diff
@@ -122,6 +123,44 @@ func (t *Torrent) Closed() <-chan struct{} {
...
+		if err != nil {
+			panic(err)
+		}
+
+		ks = append(ks, Peer{
+			Id: conn.PeerID,
+			IP: ip,
+			Port: port,
+			Source: conn.Discovery,
+			// TODO: the connection can be unencrypted due to our (or the peer's) preference,
```
anacrolix: SupportsEncryption is advisory only. I don't believe anything actually acts on it, but the information is provided by some sources.

boramalper: > but the information is provided by some sources

In this case, by which sources? `conn.encrypted` is enough or do I have another option?

torrent.go
boramalper: But how do I get their peer ID, *source*, and whether or not they support encryption? `Torrent.halfOpen` just have the hostport strings right?

anacrolix: That's right. I don't think we can as it is. We could however carry over the values when they're pulled from the pending peers list, and put it in the half-open map. Once the connection is successful, we can apply them there.

anacrolix: See b39df82fbc9269c69257ac22dea87338a886faf6

torrent.go
anacrolix: If the connection is encrypted, that's certainly enough to set SupportsEncryption. But if we're not connected to them with an encrypted connection, I couldn't say what's appropriate. We can carry forward the SupportsEncryption value as we received it from trackers/DHT/PEX, or just use the encryption state for the connection. It's probably easiest to do the latter for now.

PR #200, opened 2017-10-23
Added trigger for updatePieceCompletion()

elgatito:
When piece is completed it's marked as complete, but when it's not needed - we remove the piece and need to update completion properly, so that if we later need again this piece - it will be downloaded.

Could not find any better way other than triggering `updatePieceCompletion()` to avoid breaking the code.

### DIFF
--- a/torrent.go
+++ b/torrent.go
@@ -1004,6 +1004,11 @@ func (t *Torrent) putPieceInclination(pi []int) {
 	pieceInclinationsPut.Add(1)
 }
 
+// Trigger updatePieceCompletion
+func (t *Torrent) UpdatePieceCompletion(piece int) {
+	t.updatePieceCompletion(piece)
+}
+
 func (t *Torrent) updatePieceCompletion(piece int) {
 	pcu := t.pieceCompleteUncached(piece)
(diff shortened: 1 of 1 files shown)


anacrolix:
It looks like you want to call this from code external to the torrent package. Have you found the piece completion to fail you in some circumstance? If you open a Reader on the Torrent and it finds that data is missing it should automatically mark the piece as missing and download it again. Also have you tested this new method with the race detector? It looks to be missing Client Mutex calls.

elgatito:
> Have you found the piece completion to fail you in some circumstance? If you open a Reader on the Torrent and it finds that data is missing it should automatically mark the piece as missing and download it again. 

It's used for memory storage. I had issues with it, when I download a piece, then reader comes over it, I truncate that piece in memory, but piece completion still thinking the piece is complete, and when I manually call updatePieceCompletion() it works fine. Maybe it's related to the fact there is another piece in that memory space.

It was a long time ago, when I created that storage. I will make another shot in rewriting the storage, then  come back with more information.

anacrolix:
Thanks for the effort. I'll take another look when you have more progress.

PR #201, opened 2017-10-28
Provide trackers specific http client

dz0ny:
Mainly follows https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779
but also disables HTTPS certificate verification.


anacrolix:
What's your justification for InsecureSkipVerify? Do you have any data on typical connect/response times for HTTP trackers to support those timeout values?

dz0ny:
Some private trackers use self signed certs. And yes those timeouts are for slow trackers.

PR #203, opened 2017-11-04
Aggressive Upload

iamacarpet:
Hello @anacrolix ,

As discussed, a more aggressive upload to match the default behaviour from libtorrent.

You suggested making it the default behaviour, so I've included a config option to turn it off.

Regards,
iamacarpet


anacrolix:
I'm not sure I even want a config option, but I see maybe others will want to retain the existing behaviour, or you may even want to flip it in your own testing and report on what works best. Let's do it.

PR #215, opened 2017-12-05
added disable_libutp build flag usage

elgatito:
allow disable_libutp flag usage for build, in case cgo is used and you don't want to use libutp.

### DIFF
--- a/utp_go.go
+++ b/utp_go.go
@@ -1,4 +1,4 @@
-// +build !cgo
+// +build !cgo disable_libutp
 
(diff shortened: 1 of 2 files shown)


anacrolix:
I wouldn't advise it but I'm sure you have your reasons.

elgatito:
Just the matter of having a choice. For example, we always use cgo, but then we don't have any choice between go-utp and libutp.

anacrolix:
:+1:

PR #219, opened 2017-12-28
adding http user-agent setters and usage

elgatito:
Currently library is not setting User-Agent for HTTP requests, so by default it's 'Go-http-client/1.1' and that causes problems with trackers that have Whitelist settings.
Change allows to change default UserAgent to anything.

### DIFF
--- a/cmd/tracker-announce/main.go
+++ b/cmd/tracker-announce/main.go
@@ -39,7 +39,7 @@ func main() {
 		ar.InfoHash = ts.InfoHash
 		for _, tier := range ts.Trackers {
 			for _, tURI := range tier {
-				resp, err := tracker.Announce(torrent.DefaultHTTPClient, tURI, &ar)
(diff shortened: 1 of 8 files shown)


themihai:
I believe this could have been done by modifying  DefaultHTTPClient client(by using a wrapper over the default Transport /RoundTripper) as well. It allows you not only to change the user agent but also implement authentication(if required). That's how golang/oauth package works.

anacrolix:
That does seem a bit nicer. What do you make of this comment?

>// RoundTrip should not modify the request, except for
>// consuming and closing the Request's Body.

Maybe if I had a Client.getHttpClient method that returned everything wrapped up including User-Agent overrides it would be much nicer.

elgatito:
No need to make authentication for announcements, these endpoints are always opened, but user-agent should reflect peer client name. So for me using overrides for http.Client is too much overcomplicated.

themihai:
>> That does seem a bit nicer. What do you make of this comment?

@anacrolix You actually make a clone/copy of the request and modify the headers to avoid race conditions. See https://github.com/golang/oauth2/blob/master/transport.go#L36

PR #224, opened 2018-01-06
Download individual files

whitepitek:
This patch allows for downloading individual files in command torrent. A new option introduced: -innerPath, which is a set of paths of files inside torrents.


anacrolix:
Did you take a look at Cmd/torrent-pick?

whitepitek:
Actually I did take a brief look at it though somehow deicided it was for something different. Now I see it is exactly for what I need.

PR #232, opened 2018-01-27
Update main.go

ghost:


anacrolix:
I believe on the clause immediately above, you can already achieve the desired result by prepending `infohash:` to your infohash.

Issue #268, opened 2018-07-23
Listen to termination signals


JyBP:
The `torrent` binary doesn't listen to any termination signals.
`torrentfs` on the other hand does listen to termination signals and properly terminates the process. It would be a small but nice feature to add.

I will make a small PR soon if no one has an issue with it.

anacrolix:
Sounds good. I'm not sure what there is to clean up on a signal.

PR #269, opened 2018-07-24
cmd/torrent: Listen to termination signals

JyBP:
As suggested in #268, a small PR to terminate the `torrent` process while downloading.

As far as cleanup goes, performing a `client.Close` does everything needed. The method is goroutine-safe and the current code will no longer block on `client.WaitAll()` but proceed to `log.Fatal("y u no complete torrents?!")` which sounds reasonable to me.

Partially downloaded files are left as they are.

### DIFF
--- a/cmd/torrent/main.go
+++ b/cmd/torrent/main.go
@@ -8,7 +8,9 @@ import (
 	"net"
 	"net/http"
 	"os"
+	"os/signal"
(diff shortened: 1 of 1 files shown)


anacrolix:
Seems reasonable. Might not build on Windows. I'll merge if it passes tests.

Issue #271, opened 2018-08-25
Logging interface


Goblinlordx:
Currently, this library is logging by default which doesn't seem very nice.  There are some places where it seems to arbitrarily [output logs](https://github.com/anacrolix/torrent/blob/2f14e5dbc017d6630fae34f26aaaf9df3e1ada07/portfwd.go#L14).  This prevents me from being able to do much with them (hide them, output them with a bit of formatting etc).  Would it be possible to have the config receive a something like a simple Logger interface or something.

I am not sure where all the places that "log" automatically are but if these are the only spots then it could be something simple like:

```go
type Logger interface {
  Printf(format string, v ...interface{})
}
```

Then you would pass something like `log.New(os.Stderr, "", log.LstdFlags)` (same thing as default `log` Logger) from default config but the user (of the library) could change out the logging as necessary.  I also see `flog` in this same file... are there multiple logging libraries used throughout this project???

**edit**
Ah yes, `flog` or `github.com/anacrolix/log` has essentially the same issue in that same file and outputs things like:
`2018-08-25 20:12:18 portfwd.go:31: discovered 0 upnp devices`

It looks like `flog` has a bit of a different interface =/.  Is there any way the logging interface can be simplified and generalized for all logging?  The way this is logging doesn't really allow me to make the logs fit the format I would want nor does it let me prevent or redirect all the logging (without interacting outside with stdout).

anacrolix:
There is some inconsistent logging going on. I haven't settled on a solution I'm happy with. There is use of "log", and "github.com/anacrolix/log", the latter through which I intend to allow python logging-style levels, handlers, formatters etc.

If you have any suggestions, or PRs to clean these up, feel free to contribute.

Goblinlordx:
Sounds good~ I will see what I can do~

yonson2:
I also ended up wanting to hide the logs outputted by the program, instead of providing a logger interface and unifying both logging libraries what I endep up doing (good enough for my use-case) was to simply ensure that those `log` calls are only called if the `Debug` property of the client is set to true.

You can see an example of how it looks like [here](https://github.com/yonson2/torrent/blob/fix/only-log-when-debug/portfwd.go), I haven't submitted a pull request as I don't know if this workaround is something that seems ok to you, I'll create it if @anacrolix is ok with it.

anacrolix:
[...]

PR #272, opened 2018-08-28
Implement generalized event and log interfaces

Goblinlordx:
Ref: https://github.com/anacrolix/torrent/issues/271
 - Add `log` package containing generalized internal (`Event`) and external (`Logger`) interfaces
 - Implement a fairly generalized and flexible set of events for usage inside package (`event`)
 - Implement a simple leveled default logger (`leveled_logger`)

This PR is more to get feedback on how you feel about this type of implementation.  I feel I made it pretty flexible.  The idea is, when the client is instantiated, the config would have a `Logger` property as it does already.  I made a `Logger` interface for this so anyone can make their own `Logger`.  The `Logger` only receives `Event`s and handles them as desired.

The simplest implementation could have a NOOP style implementation of `Log` that would never log anything.  A more complex version could introspect the event and do it's own filtering and formatting as desired.

`Event` is primarily for internal implementation.  Types (which need to log something) can have a copy of the `Logger` passed to them when they are instantiated.  After that they can easily call the `Logger` with an `Event` they created.

I added a few basic `Event` types to do some basic event logging and error logging.

Below is an example usage of the `event` and `leveled_logger` subpackages:
```
package main

import (
	"github.com/anacrolix/torrent/event"
	"github.com/anacrolix/torrent/leveled_logger"
)

func main() {
	l := leveled_logger.NewLeveledLogger(event.DebugLevel)
	l.Log(event.Debug("Test2"))
	l.Log(event.Info("test"))
	l.Log(event.Error("test"))
}
```

### DIFF
--- a/event/base_event.go
+++ b/event/base_event.go
@@ -1,0 +1,52 @@
+package event
+
+import (
+	"sync"
+
+	"github.com/anacrolix/torrent/log"
+)
+
+// BaseEvent is an internal event type used for log
+type BaseEvent struct {
+	*sync.Mutex
+	data *map[string]interface{}
+}
+
+func eventMsg(e log.Event) string {
+	d := *e.Data()
+
+	var ifc interface{}
+	var msg string
+	var ok bool
+	if ifc, ok = d["msg"]; ok {
+		msg, ok = ifc.(string)
+	}
+	if !ok {
+		msg = ""
+	}
+
+	return msg
+}
+
+// String Base string method reeturns msg
+func (e BaseEvent) String() string {
+	return buildMsg(e)
+}
+
+// Data returns map of data associated with event
+func (e BaseEvent) Data() *map[string]interface{} {
+	return e.data
+}
+
+// AddValue adds a KV pair of data to the event (prevent adding of msg or level)
+func (e BaseEvent) AddValue(k string, v interface{}) {
+	if k == "msg" {
+		return
+	}
+
+	e.Lock()
+	defer e.Unlock()
+
+	d := *e.data
+	d[k] = v
+}
--- a/event/error_event.go
+++ b/event/error_event.go
@@ -1,0 +1,60 @@
+package event
+
+import (
+	"fmt"
+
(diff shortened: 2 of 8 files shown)


anacrolix:
I don't really want an logger implemented in the torrent repo unless it has behaviour required specific to the repo. The level/filter stuff I did with github.com/anacrolix/log, perhaps you can look there and see if that meets your needs?

Goblinlordx:
Logger is just to adapt events (internal interface) to an external interface.  The Event implementation was needed to be able to pass level information (which is known at call site) through the interface without the Logger needing to know implementation details of the Event (it doesn't need to know levels were used to tag Events).

The Logger interface is what does the actual logging.  `leveled_logger` is the only thing doing actual logging.  Is that the only part you are referring to not having?

Also, I think ideally, none of this would be necessary and errors would be returned where necessary or events would happen via a provided callback.  Essentially, the Logger interface I added provides that call back mechanism (which doesn't currently exist).

If a logger is provided without a simple interface (like is currently done) that means anyone providing there own needs to have a very large API surface to cover.  The idea is that internally everything would only interact with the Logger (inside this library) via the very simple Logger interface.

The log package you mentioned is fine and can be used inside the default logger but the Logger is required to prevent implementation details (larger than the interface) from leaking inside the library.

Goblinlordx:
To be clearer, only Event and Logger interfaces would be used internally.  This allows the internal implementation to add level information without the Logger needing to handle or even know that the levels exist.  The Logger can also be implemented without any knowledge of how Events are implemented (or if they have any associated data with them like level information.

Goblinlordx:
All that said, `event` does not necessarily need to be part of this library but the `log` implementation you mentioned doesn't fit the need for having a small API surface for anyone to provide their own logger.

anacrolix:
I think the idea is good, the logging in this repo (and possibly across most of my repos) is inadequate, but I haven't settled on a solution, and I don't think this is the one I want. Thanks for the contribution.

PR #286, opened 2018-10-30
reverting usage of proxy for http requests

elgatito:
Trying to revert usage of proxies for HTTP requests.
I remember you did not want to use this way, but I don't see any good approach with less actions to do.

The change will add HTTPProxy, which can differ from listener's proxy, if needed. 
Tried on my environment, looks fine, all connections are made to the proxy.

### DIFF
--- a/client.go
+++ b/client.go
@@ -10,6 +10,8 @@ import (
 	"fmt"
 	"io"
 	"net"
+	"net/http"
+	"net/url"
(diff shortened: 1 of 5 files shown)


anacrolix:
Sounds good, sorry for the run around.

anacrolix:
Very nicely done.

PR #287, opened 2018-11-03
adding custom net.Conn struct to return proper RemoteAddr for proxied…

elgatito:
… connections.

Have added custom struct to return proper RemoteAddr for proxied connections.
Built-in connections do not return it properly.

Needs some overview. I have also modified dialers composition.

PS: While debugging this, I have noticed we try each address with 4 protocols, for example, "93.85.33.185:62673" is sent to `socket.dial()` for all networks (tcp4, tcp6, udp4, udp6). Should we really fire those connections is address is not an IPv6?

### DIFF
--- a/socket.go
+++ b/socket.go
@@ -22,6 +22,48 @@ type socket interface {
 	dialer
 }
 
+type proxyConn struct {
+	net.Conn
+	remoteAddr net.Addr
+}
+
+// RemoteAddr returns real RemoteAddr of a proxied connection,
+// instead of the first hop's RemoteAddr, which is always the proxy address.
+func (c *proxyConn) RemoteAddr() net.Addr {
+	if c.remoteAddr != nil {
+		return c.remoteAddr
+	}
+
+	return c.Conn.RemoteAddr()
+}
+
+func getNetworkDialer(network string, dialer proxy.Dialer) func(ctx context.Context, network string, addr string) (net.Conn, error) {
+	return func(ctx context.Context, network string, addr string) (net.Conn, error) {
+		conn, err := dialer.Dial(network, addr)
+		if err != nil {
+			return nil, err
+		}
+
+		// When 'network' is tcp6/udp6 and address is tcp4 - we get resolve exception.
+		// As a workaround - we can run same resolve with tcp4/udp4, instead of tcp6/udp6.
+		// Currently we leave old RemoteAddr().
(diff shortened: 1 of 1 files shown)


anacrolix:
Can you describe the issues caused by having the wrong remote address? What would you think of having remoteAddr on connection instead (instead of adding the netConn wrapper), and using that everywhere that wants to do checks? I am guessing the main issue with RemoteAddr as it is, is that connections to the same proxy are handled like connections to the same peer?

I believe it's possible to connect to IPv4 addresses over IPv6, so I'm leaning toward yes, unless there's a measurable performance impact.

anacrolix:
How do incoming connections work when you have proxies? Do you just not receive any?

anacrolix:
I was thinking something like https://github.com/anacrolix/torrent/pull/288. Let me know what you think, and if it works for you.

elgatito:
> Can you describe the issues caused by having the wrong remote address? 

First of all, all connections are treated like the same peer, stats show like "10 connections out of 1 peers", because remoteaddr point to the same peer.
Not sure it was caused by that, but suddenly all connections were closed.

> What would you think of having remoteAddr on connection instead (instead of adding the netConn wrapper), and using that everywhere that wants to do checks? I am guessing the main issue with RemoteAddr as it is, is that connections to the same proxy are handled like connections to the same peer?

Probably. Not sure, but it can be.

> I believe it's possible to connect to IPv4 addresses over IPv6, so I'm leaning toward yes, unless there's a measurable performance impact.

https://en.wikipedia.org/wiki/6to4
Probably. My concern is about making 4 outgoing connections for every peer. Do they come over rate limiting? I mean, if we want to make 50 peers, do we try 50*4 or we try 50 outgoing connections?


> How do incoming connections work when you have proxies? Do you just not receive any?

They don't come from proxy, they are only proxying connections from my side. Incoming connections are working when using VPN. That is fine.

anacrolix:
> Probably. My concern is about making 4 outgoing connections for every peer. Do they come over rate limiting? I mean, if we want to make 50 peers, do we try 50*4 or we try 50 outgoing connections?
It's 50*4, for now.
> 
> > How do incoming connections work when you have proxies? Do you just not receive any?
> 
> They don't come from proxy, they are only proxying connections from my side. Incoming connections are working when using VPN. That is fine.
Okay thanks.

PR #288, opened 2018-11-04
Store network and remote address per connection

anacrolix:
Don't assume the net.Conn has the correct values.

### DIFF
--- a/bep40.go
+++ b/bep40.go
@@ -13,11 +13,6 @@ var table = crc32.MakeTable(crc32.Castagnoli)
 
 type peerPriority = uint32
 
-type ipPort struct {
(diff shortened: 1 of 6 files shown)


elgatito:
Don't know why (can't see changes to rate limiting or connections loop), but with proxy this branch is much better than mine or master.

IPs of remote connections look fine.

anacrolix:
Okay, it's possible there's system calls in calling net.Conn.RemoteAddr sometimes. This PR would avoid that cost, and also stores the IP and port in a more efficient manner. If it works for you, should I merge this over your PR?

elgatito:
Sure, go on. I will make more testing.

PR #308, opened 2019-01-13
fix: only log errors when debug is set to true

yonson2:
This is related to #271, simply put when `Debug` is set to `false` this PR makes it to the user will see almost no log calls, allowing to use this package as a third-party library and having more control of what is sent to `Stderr` or not

@anacrolix are there any log calls that you wouldn't skip showing here?


anacrolix:
I've pushed anacrolix/log throughout the codebase, and made it easier there to identify "levels" in the messages. That invalidates this PR, but I'm still interested in making the logging quieter. We can revisit this since those changes.

PR #311, opened 2019-02-06
Adds more flags to torrent-create

masihyeganeh:
Flags for setting Comment and CreatedBy fields and starting with empty AnnounceList


anacrolix:
Thanks! What did you think of `tagflag`?

PR #315, opened 2019-04-11
add one downstream project

anatasluo:
[ANT Downloader](https://github.com/anatasluo/ant) is a bittorrent client developed by golang, angular and electron. It has integrated many downsteam projects from anacrolix/torrent like goTorrent and Go Peerflix and some other public torrent services like [itorrents](https://itorrents.org/), [trackerslist](https://github.com/ngosang/trackerslist). It is a client for all platforms since golang and electron can be built in all platforms, mainly for linux since I think utorrent is not so beautiful ：D. Hope you like it.


anacrolix:
@anatasluo I think you got the link wrong. I've added a fix, can you confirm?

anatasluo:
@anacrolix Oh, it is my fault. Now, it is right, thanks~

anatasluo:
I remember I have checked it many times. It is really strange since I should not make such a mistake. Thanks for your time and patience again.

PR #318, opened 2019-04-16
torrentfs: some pathname-related fixes

rgeorgiev583:
* Fix a bug where `ENOENT` is returned when the node for an entry in the root directory of a torrent is requested.  The problem was that the full pathname to each entry in that directory would wrongly get prefixed with a slash, which means that all comparisons against the pathnames of `(*torrent).File` objects would fail as the latter are always relative.
* Fix a bug where the basenames of files are extracted incorrectly when listing the entries in a directory.  The problem was that the `dirNode.path` string was treated as a list of path components when in fact it's just a pathname string.

### DIFF
--- a/fs/torrentfs.go
+++ b/fs/torrentfs.go
@@ -102,7 +102,12 @@ func (dn dirNode) ReadDirAll(ctx context.Context) (des []fuse.Dirent, err error)
 func (dn dirNode) Lookup(_ context.Context, name string) (fusefs.Node, error) {
 	dir := false
 	var file *torrent.File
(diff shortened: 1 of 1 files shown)


anacrolix:
Thanks for finding this. Are you interested in adding a few tests?

anacrolix:
Were you able to get panics from deeply nested directories due to the bad indexing [here](https://github.com/anacrolix/torrent/pull/318/files#diff-167920e1d2ec48c84d4a6f8109975ae6L84)?

PR #321, opened 2019-05-31
Adding Started and Stopped events

martelli:
Some trackers (chihaya, https://bit.ly/2KdOu2G) require the "Stopped" event to be sent, so the seeder can be removed from the swarm. This patch adds the "Started" and "Stopped" events.


martelli:
Checking CircleCI failure.

anacrolix:
LGTM, thanks.

anacrolix:
Looks like this may have introduced some announces that occur while holding the client lock. I will find a fix shortly.

PR #322, opened 2019-07-25
fixed utf-8 names bug

powerand:


anacrolix:
Thanks for isolating the fix. See https://github.com/anacrolix/torrent/pull/323#issuecomment-521091986. I'll leave this one open for feedback.

PR #323, opened 2019-07-25
Utf8 names bug

powerand:


anacrolix:
I can't accept this PR. You may want to look into using Go modules to avoid having to rewrite your import paths like this. Can you link to or identify the UTF8 issue you're having with the name field in your metainfos? Thanks.

PR #325, opened 2019-08-08
Expose peer piece availability

josiah2009:
This will allow clients to read information relating to piece scarcity. One use case would be adjusting piece priority to download the uncommon pieces first.

### DIFF
--- a/connection.go
+++ b/connection.go
@@ -326,6 +326,29 @@ func (cn *connection) PeerHasPiece(piece pieceIndex) bool {
 	return cn.peerSentHaveAll || cn.peerPieces.Contains(bitmap.BitIndex(piece))
 }
 
+func (cn *connection) PeerBitfield() []bool {
+	if !cn.t.haveInfo() {
+		return nil
+	}
+
+	bf := make([]bool, cn.t.numPieces())
+
+	if cn.peerSentHaveAll {
+		for i, _ := range(bf) {
+			bf[i] = true
+		}
+
+		return bf
+	}
+
+	cn.peerPieces.IterTyped(func(piece int) bool {
+		bf[piece] = true
+		return true
+	})
+
+	return bf
+}
+
 // Writes a message into the write buffer.
 func (cn *connection) Post(msg pp.Message) {
 	torrent.Add(fmt.Sprintf("messages posted of type %s", msg.Type.String()), 1)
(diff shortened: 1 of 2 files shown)


anacrolix:
Looks good, can you link to where you're using the new API elements?

connection.go
```diff
@@ -326,6 +326,29 @@ func (cn *connection) PeerHasPiece(piece pieceIndex) bool {
 	return cn.peerSentHaveAll || cn.peerPieces.Contains(bitmap.BitIndex(piece))
 }
 
+func (cn *connection) PeerBitfield() []bool {
```
anacrolix: The allocation is unnecessary, perhaps make it unexported to allow changes in the future if it becomes a performance bottleneck?

torrent.go
```diff
@@ -710,6 +710,32 @@ func (t *Torrent) bitfield() (bf []bool) {
 	return
 }
 
+func (t *Torrent) CompletedPiecesBitfield() []bool {
```
anacrolix: Can't you already do this by iterating over the Torrent.Pieces and getting their status individually?

torrent.go
```diff
@@ -710,6 +710,32 @@ func (t *Torrent) bitfield() (bf []bool) {
...
+	return t.bitfield()
+}
+
+// Returns a list containing the number of peers that have a specific piece available
+func (t *Torrent) PieceAvailabilityList() (connCount int, connBitfieldSum []int) {
```
anacrolix: Perhaps we should just expose the Torrent connections list, and let people do this kind of thing on their own?

torrent.go
josiah2009: You can do that to check if a local piece progress but there is no exposed method or way to retrieve the remote status of a piece

torrent.go
josiah2009: I think that would be a good idea as currently there is little to no insight into specific peer status from an external library

josiah2009:
I am in the early stages of trying to build my own torrent client based on your library. The initial testing code is [here](https://github.com/josiah2009/storrent). My use case is going to be what I described before, implement an algorithm to prioritize the pieces with Piece.SetPriority based on if the piece is rare or not

anacrolix:
Thanks for the info. I see you're attempting to get at the request strategy, it may be easier if this were exposed as an interface that you can implement. Your plan to create an algorithm based on piece rarity will work, but won't be integrated synchronously into the decision-making that the client makes. Feel free to open any issues to expose things, and to work on exposing the request strategy more effectively.

anacrolix:
Definitely I think it would be preferable to expose methods to allow getting at the connections and their state, rather than expose a method just for piece availability alone.

PR #334, opened 2019-09-14
Improve listen all

james-lawrence:
the library regressed when providing IP addresses for the listen address. this PR adds a test and fixes the regression.

also contains misc gofmt/vet changes.

```
func TestClientIPHost(t *testing.T) {
	c := TestingConfig()
	c.ListenHost = func(string) string { return "127.0.0.1" }
	cl, err := NewClient(c)
	require.NoError(t, err)
	cl.Close()
}
```
results in 
```
    client_test.go:60: 
        	Error Trace:	client_test.go:60
        	Error:      	Received unexpected error:
        	            	listen tcp6: address 127.0.0.1: no suitable address found
        	            	subsequent listen
        	            	github.com/anacrolix/torrent.listenAllRetry
        	            		/shared/james-lawrence/development/torrent/src/github.com/anacrolix/torrent/socket.go:126
        	            	github.com/anacrolix/torrent.listenAll
        	            		/shared/james-lawrence/development/torrent/src/github.com/anacrolix/torrent/socket.go:93
        	            	github.com/anacrolix/torrent.NewClient
        	            		/shared/james-lawrence/development/torrent/src/github.com/anacrolix/torrent/client.go:230
        	            	github.com/anacrolix/torrent.TestClientIPHost
        	            		/shared/james-lawrence/development/torrent/src/github.com/anacrolix/torrent/client_test.go:59
        	            	testing.tRunner
        	            		/usr/lib/go/src/testing/testing.go:909
        	            	runtime.goexit
        	            		/usr/lib/go/src/runtime/asm_amd64.s:1357
        	Test:       	TestClientIPHost
```

problem is a hostname isn't necessarily available for the given network you want to listen too, nor is the availability IPv4 or 6. this PR removes those assumptions.


anacrolix:
The IPv4/IPv6 differentiation is required for using DHT correctly on different networks. The ListenHost config field is changed because as you mention, IPs differ between the networks, particularly when you listen on them separately. Could you explain the specific scenario that doesn't work with the existing ListenHost config field?

anacrolix:
Please describe that case in a new issue.

james-lawrence:
> Could you explain the specific scenario that doesn't work with the existing ListenHost config field?

Sorry thought I did through the test. if you assign a IP address for the network interface you want to listen on the library client fails unnecessarily, yes one could set the disabled flags for IPv4/6 but those fields shouldn't even exist (in fact this PR should allow them to be removed, simplifying configuration).

> The IPv4/IPv6 differentiation is required for using DHT correctly on different networks (tcp4 vs tcp6 vs tcp for example)

please explain this further, because these changes functionally change nothing in terms of what what interfaces would be listened on for a particular hostname, nor what is passed around in the library to the DHT.

at the end of the day `s, err := listen(nah.Network, net.JoinHostPort(nah.Host, portStr), proxyURL, f)` still does the exact same thing `lookupIP` does, namely resolve the host to a set of IP addresses. only thing the library does here is over complicate itself by specifically forcing listen to a particular network for the provided hostname instead of just looking up the IPs for that hostname.

james-lawrence:
alternatively, switching to a `client.Serve(net.Listener)` model with a Helper `AutoServe(client)` would be easier to use and configure by end users.

elgatito:
@anacrolix  as far, as I remember, split of a ListenAddr was added, because IPv4 and IPv6 addresses differ, and it can be different for each use case.
Anyway, anacrolix/torrent is a library for other projects, so removing configuration part, for simplifying someone's usage, does not look good for me.

anacrolix:
> if you assign a IP address for the network interface you want to listen on the library client fails unnecessarily, yes one could set the disabled flags for IPv4/6

The tests don't show behaviour that currently isn't possible with the existing configuration. If IPv6 or IPv4 listens fail, it may be an improvement to not fail the listenAll, and that would be a useful thing to test.

> please explain this further, because these changes functionally change nothing in terms of what what interfaces would be listened on for a particular hostname, nor what is passed around in the library to the DHT

DHT works differently depending on whether the network is on IPv6 or IPv4. Rather than guess from the IP addresses, which aren't strictly disjoint between those networks, and are clouded by dualstack handling, we separate the networks entirely because the behaviour must differ.

> only thing the library does here is over complicate itself by specifically forcing listen to a particular network for the provided hostname instead of just looking up the IPs for that hostname

It's worth pointing out that it's not a requirement that connections even be over IP. If you want to do your own lookups, you can with the current configuration-style, and that is intentional. As @elgatito points out, flexibility has a higher priority than convenience here. You could create a wrapper if you want to wave this stuff away.

james-lawrence:
just jumping back to this sorry. =) so lets me start by saying: I'm not wedded to these changes as they stand. in fact I wrote them to minimize in this way to minimize changes to the library API. My prefer solution would be a single method on the client:

```golang
func (t Client) Bind(l net.Listener) error {
   // pseudo code
   d, err := buildDHT(l)
   if err != nil {
      return err
   }

   t.m.Lock()
   defer t.m.Unlock()
   t.dhts = append(t.dhts, d)
   return nil
}

func Autolisten(c Client, hostnames ...string) error {
   // perform current behaviour.
}
```

which is by and far the most flexible option, while simultaneously reducing code within this library:
1. allows network stacks to be provided that are completely unrelated to IP.
2. allows removal of all the configuration options are network handling from this library.
3. allows multiple networks to be listened to simultaneously (without listening to 0.0.0.0)

if the above is acceptable then I'm happy to do that work instead. if that is not acceptable, then everything below still probably applies.

> The tests don't show behaviour that currently isn't possible with the existing configuration. 

I didn't say the behaviour wasn't possible. I'm saying its a regression and very unexpected. I wouldn't have expected to have to write custom code to bind to a literal IP network address. There are far easier, less surprising, and more idiomatic ways to do this.

> DHT works differently depending on whether the network is on IPv6 or IPv4...
> we separate the networks entirely because the behaviour must differ.

I understand that. its also irrelevant to these changes.

> Rather than guess from the IP addresses, which aren't strictly disjoint between those networks, and are clouded by dualstack handling

there is no guessing of IP addresses in these changes (unless you mean the IPv6 -> IPv4 resolution, which only exists because the current code does that). the current code ***is*** actually guessing. for example on the CI ubuntu host 'localhost' doesn't actually have an IPv6 address associated with it. current code still attempts to connect to IPv6. these changes don't.

on a system where localhost is configured properly for dual stack, both ipv4 and ipv6 addresses are returned and used.

> It's worth pointing out that it's not a requirement that connections even be over IP

Agreed! but that actually *is* a requirement in the current code, and the PR. (my alternative I suggestion removes this restriction).

> If you want to do your own lookups.

I don't, I want to do no hostname lookups, which is why I was passing it the network IP address directly. which then promptly broke, as clearly demonstrated by the test. this is a regression from previous versions. use to work just fine.

> As @elgatito points out, flexibility has a higher priority than convenience here. You could create a wrapper if you want to wave this stuff away.

the current code and the PR are almost identical in in terms of flexibility. 
1. both require that any hostname resolves to a single IP address.
2. both only allow for a single hostname to be resolved.
3. both allow you to be foolish and write your own hostname -> ip resolution scheme.
4. both will detect and use both IPv4 and IPv6 addresses for a hostname with a properly configured system.

show me a situation where it doesn't match the current behaviour and I'll likely be able to show you a misconfigured system.

james-lawrence:
@elgatito just an FYI as i attempted to explain, the reduction in configuration is because it is no longer necessary. there is no reduction in functionality. this library is simply re-inventing the wheel poorly and as a result requires additional flags to work properly.

elgatito:
@james-lawrence Show me how can I set different listen hosts for IPv4 and IPv6 with your changes.

james-lawrence:
@elgatito I'd ask you why you have two different hostnames for the same system. I'd also question your use of the term hostname. I believe you mean IP addresses?

but here is how you'd configure different addresses for a single hostname.
/etc/hosts:
```
::2 example.com
127.0.0.1 example.com
```

```
package main

import "net"
import "log"

func main() {
	host := "example.com"
	ips, err := net.LookupIP(host)
	if err != nil {
		log.Printf("host lookup failed \"%s\": %s\n", host, err)
	}
	fmt.Println("ips", ips)
}
// Output: ips [127.0.0.1 ::2]
```

anacrolix:
In the example: `c.ListenHost = func(string) string { return "127.0.0.1" }`, you should not be ignoring the network parameter. See https://github.com/anacrolix/torrent/blob/3ba991e38aa4d0019e50c05df79651c0d16f3003/listen.go#L5-L11 for an example of this. The API may be unfortunate in some regards, the real win here would be to not have the client do the listening for you (here it's catering to common users and use cases). You [described a method to add user-provided sockets](https://github.com/anacrolix/torrent/pull/334#issuecomment-544245387), the internals have partly moved toward this. I think this is a better direction to head in. As it stands, to the best of my knowledge, this PR will negatively impact the way DHTs are currently initialized, and I don't see a clear advantage to the change to network handling.

james-lawrence:
fair enough, shall i generate a PR for the alternate approach i described?

anacrolix:
Yes

anacrolix:
I'm closing this in favour of smaller, more specific PRs that may come.

PR #339, opened 2019-10-27
Add logging level in upnp discovery

davtoro:
add missing logging level, otherwise it shows always


davtoro:
Ah, sorry, I didn't notice that you've added option to discard logging in newer version.

anacrolix:
Thanks. Let me know if there's any inadequacies in the new logging system.

PR #343, opened 2019-11-11
Export Peer function

anthonyraymond:
The method `fromDictInterface` was not exported from the `Peer` structure.

People using this `Peer` struct from outside the library would be happy to access the method to instantiate peers from a non-compact format.
This method is a constructor in some way, it should be exported to provide users a convenient way to instantiate a peers from a map without having to rewrite the same logic

### DIFF
--- a/tracker/http.go
+++ b/tracker/http.go
@@ -56,7 +56,7 @@ func (me *Peers) UnmarshalBencode(b []byte) (err error) {
 		vars.Add("http responses with list peers", 1)
 		for _, i := range v {
 			var p Peer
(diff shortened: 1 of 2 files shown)


anacrolix:
Can you tell me where you're using this?

anthonyraymond:
Hi @anacrolix, i'd like to use `Peer{}.fromDictInterface()` while parsing a tracker response. It's actually pretty similar to the way you are using it. With the difference that you can use this unexported method from your own package 😄.
Since the bdecoded dictionary is a normalized format i'd like this method to be exposed.

Issue #341, opened 2019-11-06
Peer lists are not exchanged over PEX


yarikk:
PEX support in the library is limited to handling incoming PEX messages and updating own peer lists. The list of currently connected peers is not being communicated.

[BEP 11](http://bittorrent.org/beps/bep_0011.html) regulates the frequency of batch updates to  happen no more than once a minute and have no more than 50 entires, but it's better be kept flexible.

We are looking into creating a tracker-less setup in an isolated managed environment, with initial seed addresses bootstrapped via metadata or magnet links.

@anacrolix Any pointers before I jump into implementing this?

anacrolix:
@yarikk that's right. I'm sure I had an issue and TODO regarding this, but I cannot find them. I welcome your contributing this. You'll probably want to add code to generate the outgoing pex message here: https://github.com/anacrolix/torrent/blob/289293f3cc0cfd03bbe3bf168dcc1ba6a90d13aa/connection.go#L590. You'll probably also want a timer like the keep alive one, to trigger updates like this one: https://github.com/anacrolix/torrent/blob/289293f3cc0cfd03bbe3bf168dcc1ba6a90d13aa/connection.go#L602. Let me know what you think, it looks like there's some interesting requirements on the implementation from reading the BEP.

yarikk:
Addressed in #341.

anacrolix:
@yarikk I wasn't sure where else to message you, I've made some recent changes to PEX to use net/netip, and some fixes to the interval use (I think it was applied for the whole torrent, rather than per peer). Do you want to take a look?

PR #353, opened 2019-12-11
Share current connections with peers over PEX (anacrolix#341)

yarikk:
Context in issue #341. Work in progress – sharing for visibility and early feedback.

I've been testing the new functionality in an isolated Docker-based environment with trackers and DHT functions disabled. The swarm consisted 1 rate-limited seeder and 7 downloaders whose peer lists were bootstrapped with the seeder's IP+port.  Even with a simple exchange of connected peers following the handshake, without the follow-up PEX updates, the swarm has grown visibility of each other and got fairly well cross-connected, with chunks download by one peer quickly propagating to the others.

The existing protocol isn't much affected other than sending an extra message with an ID negotiated according to [BEP 10](http://bittorrent.org/beps/bep_0010.html). This may potentially result in interoperability issues; with that said, if encountered in field, the clients can work around by setting *DisablePEX* configuration option.

### DIFF
--- a/client.go
+++ b/client.go
@@ -934,22 +934,6 @@ func (cl *Client) sendInitialMessages(conn *PeerConn, torrent *Torrent) {
 	}
 }
 
-func (cl *Client) sendInitialPEX(conn *PeerConn, t *Torrent) {
-	xid, ok := conn.PeerExtensionIDs[pp.ExtensionNamePex]
-	if !ok {
-		return
-	}
-	m, seq := t.pex.Genmsg(0)
-	conn.pexSeq = seq
-	if m.Len() == 0 {
-		cl.logger.Printf("no initial PEX this time")
-		// FIXME see how can we schedule another initial for later
-		return
-	}
-	conn.logger.Printf("sending initial PEX message: %v", m)
-	conn.post(m.Message(xid))
-}
-
 func (cl *Client) dhtPort() (ret uint16) {
 	cl.eachDhtServer(func(s DhtServer) {
 		ret = uint16(missinggo.AddrPort(s.Addr()))
--- a/client.go
+++ b/client.go
@@ -935,21 +935,19 @@ func (cl *Client) sendInitialMessages(conn *PeerConn, torrent *Torrent) {
 }
 
 func (cl *Client) sendInitialPEX(conn *PeerConn, t *Torrent) {
-	peerPexExtendedId, ok := conn.PeerExtensionIDs[pp.ExtensionNamePex]
+	xid, ok := conn.PeerExtensionIDs[pp.ExtensionNamePex]
 	if !ok {
-		// peer did not advertise support for the PEX extension
-		conn.logger.Printf("no PEX support -  not sending initial")
 		return
 	}
-	pexMsg := t.pexInitial()
-	if pexMsg == nil {
-		// not enough peers to share — e.g. len(t.conns < 50)
-		conn.logger.Printf("skipping PEX initial")
+	m, seq := t.pex.Genmsg(0)
+	conn.pexSeq = seq
+	if m.Len() == 0 {
+		cl.logger.Printf("no initial PEX this time")
+		// FIXME see how can we schedule another initial for later
 		return
 	}
-	log.Printf("preparing PEX initial message: %v", pexMsg)
-	tx := pexMsg.Message(peerPexExtendedId)
-	conn.post(tx)
+	conn.logger.Printf("sending initial PEX message: %v", m)
+	conn.post(m.Message(xid))
 }
 
 func (cl *Client) dhtPort() (ret uint16) {
--- a/client.go
+++ b/client.go
@@ -934,6 +934,24 @@ func (cl *Client) sendInitialMessages(conn *PeerConn, torrent *Torrent) {
 	}
 }
 
+func (cl *Client) sendInitialPEX(conn *PeerConn, t *Torrent) {
+	peerPexExtendedId, ok := conn.PeerExtensionIDs[pp.ExtensionNamePex]
+	if !ok {
+		// peer did not advertise support for the PEX extension
+		conn.logger.Printf("no PEX support -  not sending initial")
+		return
+	}
+	pexMsg := t.pexInitial()
+	if pexMsg == nil {
+		// not enough peers to share — e.g. len(t.conns < 50)
+		conn.logger.Printf("skipping PEX initial")
+		return
+	}
+	log.Printf("preparing PEX initial message: %v", pexMsg)
+	tx := pexMsg.Message(peerPexExtendedId)
+	conn.post(tx)
+}
+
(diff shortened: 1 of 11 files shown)


client.go
```diff
@@ -933,6 +933,24 @@ func (cl *Client) sendInitialMessages(conn *connection, torrent *Torrent) {
...
+func (cl *Client) sendInitialPEX(conn *connection, t *Torrent) {
+	peerPexExtendedId, ok := conn.PeerExtensionIDs[pp.ExtensionNamePex]
+	if !ok {
+		// peer did not advertise support for the PEX extension
+		conn.logger.Printf("no PEX support -  not sending initial")
+		return
+	}
+	pexMsg := t.pexInitial()
+	if pexMsg == nil {
+		// not enough peers to share — e.g. len(t.conns < 50)
```
anacrolix: This doesn't match the check in `pexInitial` anymore?

connection.go
```diff
@@ -87,11 +87,12 @@ type connection struct {
...
-	PeerInterested     bool
-	PeerChoked         bool
-	PeerRequests       map[request]struct{}
-	PeerExtensionBytes pp.PeerExtensionBits
+	PeerID                PeerID
+	PeerInterested        bool
+	PeerChoked            bool
+	PeerRequests          map[request]struct{}
+	PeerExtensionBytes    pp.PeerExtensionBits
+	PeerPrefersEncryption bool // as indicated by 'e' field in extension handshake
```
anacrolix: I'm guessing you've checked this isn't recorded elsewhere. Currently I guess it's not come up that we may want to remember the prefer encryption flag, we just use it in selecting peers and then forget about it, but now we're propagating this information at a point after connecting?

anacrolix:
It looks like the test failure is just the torrentfs/FUSE stuff being flaky.

client.go
yarikk: Well, it's oversimplification, and not exactly precise. The spec says if it's less that 25, we've got to add a bunch of last known good peers from the log, 25 max, for "good" defined as fully hand-shaken connections, for each of ip4 and ip6. But if there is not enough context, the initial can be delayed. Hereby the nil is the way to tell the caller it's not a good time.

connection.go
yarikk: Well, the preference is communicated in the extended handshake, and the PEX spec tells us to pass the word so they can better decide whom to dial. So indeed we keep that indication around just for sharing with peers, and by far it was useless to keep, but that's a fairly low overhead. And we, as receivers of said flags, may also want to match peers against our preference when picking a peer to contact next.

yarikk:
@anacrolix, this is ready for review.

yarikk:
(The last push was to fix pollution of "dropped" lists. That was caused by connections which were dropped pre-maturely being recorded on the PEX feed without ever being added)

anacrolix:
Trying it out now. So far it looks great. Thanks for the effort!

pexconn.go
```diff
@@ -0,0 +1,116 @@
...
+	return tx
+}
+
+// Share is called from the writer goroutine if when it is woken up with the write buffers empty
+func (s *pexConnState) Share(postfn messageWriter) {
+	select {
+	case <-s.gate:
+		if tx := s.genmsg(); tx != nil {
+			s.dbg.Print("sending PEX message: ", tx)
+			postfn(tx.Message(s.xid))
```
anacrolix, line 76: I think you might need to return the bool here, in case there are further write attempts to the post function, even though at present sharing PEX is the last thing checked.

pexconn.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+// per-connection PEX state
+type pexConnState struct {
+	enabled bool
+	xid     pp.ExtensionNumber
+	seq     int
+	timer   *time.Timer
+	gate    chan struct{}
+	readyfn func()
```
anacrolix, line 24: 👍

pexconn.go
```diff
@@ -0,0 +1,116 @@
...
+	pexInterval   = 1 * time.Minute
+)
+
+// per-connection PEX state
+type pexConnState struct {
+	enabled bool
+	xid     pp.ExtensionNumber
+	seq     int
+	timer   *time.Timer
+	gate    chan struct{}
```
anacrolix, line 23: Anyway to combine this with the timer, or is it not worth it? I had a look and it seems AfterFunc doesn't use Timer.C, and your way saves a goroutine, so is that preferable?

torrent.go
```diff
@@ -1200,6 +1204,9 @@ func (t *Torrent) deleteConnection(c *PeerConn) (ret bool) {
 	}
 	_, ret = t.conns[c]
 	delete(t.conns, c)
+	if !t.cl.config.DisablePEX {
+		t.pex.Drop(c)
```
anacrolix, line 1208: Perhaps pex could be an interface, and a dummy provided when PEX is disabled. That way we're not testing everywhere whether we should be calling methods on it.

pexconn.go
anacrolix: It looks like uploads occur after PEX, so we do want to handle this I think.

pexconn.go
```diff
@@ -0,0 +1,116 @@
...
+	readyfn func()
+	torrent *Torrent
```
anacrolix, line 25: Possibly put this behind an interface?

anacrolix:
I've merged, please respond to my comments anyway if you can.

torrent.go
yarikk: It could, yes, but without a good deal of an alternative behaviour this doesn't feel like a big win. Also, we could have these checks internally in the methods but there are already quite a bit of things going on already.  I think what we've got here is a good balance on keeping clutter out.

pexconn.go
yarikk: That had crossed my mind, too: it was one of the things I tried. It comes out, however, that one can't use Timer.C with AfterFunc, getting a deadlock: apparently, AfterFunc consumes the channel internally. Introduction of this gating channel had helped to resolve the data races flagged by the race detector, lock-less.

pexconn.go
yarikk: Glad you like it 🙂

pexconn.go
yarikk: Good point. Let me take care of this.

pexconn.go
yarikk: Without another meaningful thing to put behind the interface, it feels like it would just add clutter without any real return. I hope it makes sense.

pexconn.go
yarikk: Here, #394.

yarikk:
@anacrolix, thanks for reviewing and merging! I answered your comments.

pexconn.go
anacrolix: It does!

anacrolix:
Thanks for following up.

PR #354, opened 2019-12-13
Possibility to change UpnP ID

davtoro:
New config option that one can set to change UpnP ID, which is listed as service name in routers etc. It's still anacrolix/torrent by default.

### DIFF
--- a/config.go
+++ b/config.go
@@ -32,6 +32,7 @@ type ClientConfig struct {
 	ListenHost              func(network string) string
 	ListenPort              int
 	NoDefaultPortForwarding bool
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -32,6 +32,7 @@ type ClientConfig struct {
 	ListenHost              func(network string) string
 	ListenPort              int
 	NoDefaultPortForwarding bool
+	UpnPID			string
```
anacrolix: Would you make it `UpnpId` or `UpnpID`?

davtoro: Sure, changed.

Issue #359, opened 2019-12-21
connection allowed even if proxy is misconfigured


sickyoon:
It appears that torrent can be retrieved through non-proxied connection if proxy_url is misconfigured.

### Steps to reproduce
To reproduce, run `cmd/torrent/main.go` with `ClientConfig.ProxyURL` set to invalid arbitrary socks5 url.

### Expected
Torrent should not download

### Actual
Torrent downloads successfully

sickyoon:
Should disable accepting connections if proxy is configured

anacrolix:
Was this fixed by https://github.com/anacrolix/torrent/pull/360?

sickyoon:
verified fixed.

PR #360, opened 2019-12-22
disable listeners if proxy is configured

sickyoon:
disable listeners if proxy is configured
https://github.com/anacrolix/torrent/issues/359

### DIFF
--- a/socket.go
+++ b/socket.go
@@ -2,6 +2,7 @@ package torrent
 
 import (
 	"context"
+	"fmt"
 	"net"
 	"net/url"
 	"strconv"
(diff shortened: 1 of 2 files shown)


sickyoon:
dependency changes (go.mod) are there for upstream project.
I will remove them once you review this PR.

socket.go
```diff
@@ -55,11 +56,14 @@ func listenTcp(network, address, proxyURL string) (s socket, err error) {
 	// If we don't need the proxy - then we should return default net.Dialer,
 	// otherwise, let's try to parse the proxyURL and return proxy.Dialer
 	if len(proxyURL) != 0 {
+
+		dl := disabledListener{l}
+
 		// TODO: The error should be propagated, as proxy may be in use for
 		// security or privacy reasons. Also just pass proxy.Dialer in from
 		// the Config.
 		if dialer, err := getProxyDialer(proxyURL); err == nil {
```
anacrolix: Wouldn't it be more appropriate instead to not ignore errors here? I'm not sure about how listening should be handled when proxies are in use.

sickyoon: Yeah we should not ignore url parse error here. My understanding is that listening can't be handled through socks5 proxy and should be disabled when proxy is in effect.

anacrolix: That makes sense. Do you want to make a PR for handling the error separately?

anacrolix: I think @elgatito would know more about this, did you want to make any recommendations, Denis?

elgatito: @anacrolix When I initially added proxy support, I've tested cases when specifying non-working proxy or proxy with wrong credentials. In both cases downloading was not working (as it should be).
Testing can by done with installing dante proxy (pretty simple to install) with authorization, and proxying through it.

sickyoon: @elgatito perhaps listeners were added after your implementation but torrent can definitely downloaded with invalid proxy. You can try to test it yourself. @anacrolix I added error handling part to my PR? could you review and merge?

anacrolix: Erroring on a bad proxy seems like the more correct behaviour, for privacy and security reasons.

Issue #356, opened 2019-12-17
TLS support


yarikk:
The RC4 cipher currently part of BitTorrent protocol encryption is considered insecure these days. In search for replacement, TLS appears the best candidate, comprising a modern, widely deployed suite of cryptographic protocols. Although there is no active or draft BEP specification on TLS, it would be a major step forward in the protocol development. The proposal  here is to  develop a not-yet-standard extension adding TLS support to BitTorrent protocol, with a perspective to have it thoroughly documented and put up as a BEP.

Any constructive input is welcome.

anacrolix:
RC4 is only used as obfuscation in BitTorrent, it's probably much easier in most cases to figure out which infohash is being used as the secret key. I'm not sure how you would change this. Can you link to any upstream proposals?

yarikk:
This is an original proposal, to be referenced in further discussions. The idea is to wrap the whole peer protocol into a TLS session, either right from the beginning or after negotiating during the basic handshake, to be decided. Therefore every peer connection is going to have own set of negotiated keys.

chantra:
I think essentially what would be useful is to be able to provide a custom dialer in https://github.com/anacrolix/torrent/blob/d34dfa6575c49eaad2dd2a6b35101f1bce26da5e/socket.go#L69, if the custom dialer is nil, it would default to net.Dialer{}. Similarly for the Listeners. 

This could be propagated through ClientConfig by using 
 something like DialerFunc, ListenerFunc which would have some default implemetation if not overriden.
https://github.com/anacrolix/torrent/blob/d34dfa6575c49eaad2dd2a6b35101f1bce26da5e/client.go#L230 would need to pass those specifically, or the whole ClientConfig.

This way, someone that want to implement customer dialer (for instance TLS), could put the logic in their own dialer, where they would need to handle validation and all on their own. For lack of existing BEP, this may give flexibility without adding complexity to the library.

This would work for the use-case we are interested in, e.g using TLS between peers, but dealing with certificates and their validation is very in-house. Solutions within the boundaries of an organizations may be pretty simple (supposedly, all leechers could use the same certs and clients would validate that cert), but TLS over public swarm would seem much more difficult to achieve.

@anacrolix any thoughts?

anacrolix:
[...]

PR #365, opened 2020-01-08
 Break out dialer and accepting listener conn 

ericsage:
Addresses #356 

During testing I found that passing in a unix domain socket doesn't work because there is an assumption that addresses have port numbers. So if that was fixed, one could use TmpDir for the test, and of course pass in domain sockets in real client code.

My first approach was to try and make `NewClient` a wrapper around `NewClientWithSockets`, but the `listenAll` dependency on `cl.firewallFunction` meant that the client has to be instantiated before socket creation. That is why I split it out into `initClient` and `initSockets`. If, in a future API, Client was more like http.Server and separated construction from starting, then the client would be ready to go before we make the sockets.

### DIFF
--- a/socket.go
+++ b/socket.go
@@ -13,13 +13,13 @@ import (
 	"golang.org/x/net/proxy"
 )
 
-type dialer interface {
+type Dialer interface {
 	dial(_ context.Context, addr string) (net.Conn, error)
 }
 
 type socket interface {
 	net.Listener
-	dialer
+	Dialer
 }
 
 func getProxyDialer(proxyURL string) (proxy.Dialer, error) {
--- a/client.go
+++ b/client.go
@@ -25,7 +25,6 @@ import (
 	"github.com/anacrolix/missinggo/pubsub"
 	"github.com/anacrolix/missinggo/slices"
 	"github.com/anacrolix/missinggo/v2"
-	"github.com/anacrolix/missinggo/v2/conntrack"
 	"github.com/anacrolix/sync"
 	"github.com/davecgh/go-spew/spew"
 	"github.com/dustin/go-humanize"
@@ -58,6 +57,7 @@ type Client struct {
 	peerID         PeerID
 	defaultStorage *storage.Client
 	onClose        []func()
+	dialers        []Dialer
 	conns          []socket
 	dhtServers     []*dht.Server
 	ipBlockList    iplist.Ranger
@@ -227,6 +227,10 @@ func NewClient(cfg *ClientConfig) (cl *Client, err error) {
 		}
 	}
 
+	if cl.config.NoDefaultConns {
+		return
+	}
+
 	cl.conns, err = listenAll(cl.listenNetworks(), cl.config.ListenHost, cl.config.ListenPort, cl.config.ProxyURL, cl.firewallCallback)
 	if err != nil {
 		return
@@ -275,6 +279,9 @@ func (cl *Client) enabledPeerNetworks() (ns []network) {
 		}
 	}
 	return
+}
+
+func AddIncomingConneciton(conn net.Conn) {
 }
 
 func (cl *Client) listenOnNetwork(n network) bool {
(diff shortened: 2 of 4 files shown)


anacrolix:
I think it would be better to provide the Sockets in the ClientConfig, or have a Client.AddSocket, and provide a ClientConfig option to disable implicit listening.

socket.go
```diff
@@ -13,13 +13,20 @@ import (
...
+	// Dial dials the peer using an implementation defined addr to etablish a
+	// stream-oriented connection with which to make client requests. Dialing
+	// must be cancelable via the provided context.
+	Dial(ctx context.Context, addr string) (net.Conn, error)
+}
+
+// Socket is a generic stream-oriented peer interface that implements
+// net.Listener to handle incoming requests from peers, and Dialer to handler
+// outgoing requests to peers.
+type Socket interface {
```
anacrolix: What about splitting this into `Listener` and `Dialer`, there are situations where we want to distinguish them (like proxies), and we can avoid juggling and guessing which is which.

anacrolix:
On further thought, why even add Listeners directly, maybe we could just add incoming streams to `Client`. There's no other behaviour on those.

ericsage:
> On further thought, why even add Listeners directly, maybe we could just add incoming streams to `Client`. There's no other behaviour on those.

What kind of API do you have in mind?

anacrolix:
@ericsage If you're keen, change `Client.conns` into `Client.dialers`. Add implicit listeners (those created automatically per the `ClientConfig`) to a hook on `Client.Close`. Do `Client.acceptConnections` on implicit listeners, as well as pass them to each DHT server as currently occurs, but don't retain them. Then add `Client.AddListener` that adds an explicit listener (without all the behaviour for the implicit listeners) and/or a `Client.AddIncomingConnection` that skips the `Accept`, and runs the accept filter logic and then starts processing the incoming connection.

Add a `Client.AddDialer` method.

To avoid having to suppress normal listening behaviour by having to specify a bunch of `DisableX` in the `ClientConfig` (and thereby disabling other `Client` features while doing it), add something to the `ClientConfig` that suppresses the normal listening behaviour.

If it's too much, I'll take what you've started in this PR, and adapt it in that direction.

james-lawrence:
this is similarly related to #334.

ericsage:
@anacrolix so if I understand correctly:
- You can either start or not start all the implicit listeners as a group.
- You can add extra listeners via AddListener.
- You can add incoming net.Conns via AddIncomingConnection and run conn acceptance in the consumer.
- Dialers are no longer coupled with the Listener protocol.

Also from my understand of the code, all the added dialers are invoked for outgoing connections are race, so if we want to do TLS, we'd only add a single dialer.

If this is correct, I can definitely do that within the next couple of days. My only fear is the number of tests I would break/have to add by changing the API.

@chantra

anacrolix:
@ericsage Yes I think you have it right. I think none of the existing tests would change, since the existing listening API would remain unchanged. It might be worth skipping AddListener, since all it would do is Accept and call AddIncomingListener, and then we have to decide how to handle errors, easier to just pass that decision to the user. I think a single ClientConfig option to disable the existing implicit listener/dialer behaviour is sufficient. That would mean you could leave `socket` in there for now, but it would only be used by the existing code, and map directly onto the new Listener (if you have one), and Dialer.

anacrolix:
@ericsage @chantra: Just checking in on this: several issues should be easier to address with the proposed changes here.

chantra:
@anacrolix I think @ericsage had a clear idea of how to proceed but other stuffs came out. @ericsage should be able to provide a status of the stack/estimate.

ericsage:
@anacrolix sorry, I've been busy with work stuff, I took a look at it tonight and came up with [this commit](https://github.com/ericsage/torrent/commit/702dadba7d1e9c474f89746b1f6f50b7980edd0c), but one issue I ran into is that with reusing the dial racing code, dialers with a corresponding Listener don't have a local addr to use. Is this along the lines of what you wanted?

anacrolix:
@ericsage Yes, although I would do away with `Client.conns` entirely. I'll put other comments on the commit directly. If you like I can take what you've started and finish implementing this, let me know.

ericsage:
@anacrolix feel free to take it over, I probably won't get around to working on this until late next week otherwise. Thanks.

anacrolix:
Ok, I'll take this on.

anacrolix:
See https://github.com/anacrolix/torrent/issues/380 and https://github.com/anacrolix/torrent/labels/custom%20networks.

PR #394, opened 2020-04-15
PEX: integrate with send throttling

yarikk:
Prevent further write attempts if the write buffer is full. Extends #353.


anacrolix:
Thanks!

PR #397, opened 2020-04-16
cmd/torrent: add -ipv4/-ipv6 options

yarikk:
Allow testing in different IP stack combinations.

### DIFF
--- a/cmd/torrent/main.go
+++ b/cmd/torrent/main.go
@@ -141,6 +141,8 @@ var flags = struct {
 	Dht             bool
 	TcpPeers        bool
 	UtpPeers        bool
+	Ipv4            bool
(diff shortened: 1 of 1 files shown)


yarikk:
Example of a run in an environment lacking ipv6:
```
% torrent download '-addr=:44444' '-uploadRate=5MB' '-dht=false' -stats -debug -seed x.torrent
2020-04-16T11:59:43+0000 NONE main.go:183: error in main: creating client: subsequent listen: listen tcp6 :44444: socket: address family not supported by protocol
```
With the new `-ipv6=false` it works just fine there.

Likewise, specifying `-ipv4=false` allows to verify behaviour in an IPv6-only environment.

PR #408, opened 2020-05-30
Add per-torrent ability to disable uploading

max-b:
Provides for ability to mark specific torrents as "don't upload".

### DIFF
--- a/peerconn.go
+++ b/peerconn.go
@@ -1325,6 +1325,9 @@ func (c *PeerConn) uploadAllowed() bool {
 	if c.t.cl.config.NoUpload {
 		return false
 	}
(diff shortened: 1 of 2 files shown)


anacrolix:
That should do it. It might be desirable to add an option when the torrent is added too. Probably to https://pkg.go.dev/github.com/anacrolix/torrent?tab=doc#Client.AddTorrentSpec.

anacrolix:
There's also some logging calls that might want to be removed.

max-b:
Thanks! I can make another quick PR for the logs and for the `AddTorrentSpec`

anacrolix:
Sounds excellent!

PR #414, opened 2020-06-19
add method AddTorrentFromUrl

HanFa:
This patch adds `AddTorrentFromUrl(torrentURl string)`.

It allows torrent to be added given its url.


anacrolix:
This is a nice convenience, but I want to move away from adding more methods to add torrents to the Client. It's not clear if the user would want to use a custom HTTP client, or specify options while adding the torrent, so I think this isn't worthwhile (even if Go's error handling is super tedious).

PR #433, opened 2020-11-06
Fix initial pex message regression

anacrolix:
As pointed out by @yarikk  https://github.com/anacrolix/torrent/commit/c1d189ed31af99cbf6f10d9ffdc2cc4e19a73b91#r43920706


yarikk (approved):
lgtm

At some point, we may want to cap the initial message so it is not overwhelmingly large.

Also, a further optimization is possible, as for every new connection the initial message has the last sent initial message as a prefix.

PR #435, opened 2020-11-09
optimise generation of the initial PEX

yarikk:
This requests pull of three commits:

* add TestPexInitialNoCutoff to catch regressions like anacrolix/torrent#433
* simplify `pexMsgFactory` to make better use of surrounding types
* optimise generation of the initial PEX

The latter deserves to be covered in greater detail.

The optimisation comes from the observation that every subsequent PEX message has to scan over the same bunch of events plus a small number of events on top of that. Included are benchmarks runnable with `go test -bench`.

Below are the benchmark results, and a plot showing the values of the last column along with the trend lines:

 * before:
```
BenchmarkPexInitial4-8     	  102764	     11319 ns/op
BenchmarkPexInitial50-8    	    2024	    500112 ns/op
BenchmarkPexInitial100-8   	     656	   1851580 ns/op
BenchmarkPexInitial200-8   	     159	   8022727 ns/op
BenchmarkPexInitial400-8   	      40	  30888163 ns/op
```

 * after:
```
BenchmarkPexInitial4-8     	  145088	      7539 ns/op
BenchmarkPexInitial50-8    	    6163	    187419 ns/op
BenchmarkPexInitial100-8   	    1693	    629353 ns/op
BenchmarkPexInitial200-8   	     465	   2236348 ns/op
BenchmarkPexInitial400-8   	     157	   7350410 ns/op
```

PEXOptimisationPlot

[PEX optimisation.pdf](https://github.com/anacrolix/torrent/files/5518647/PEX.optimisation.pdf)

### DIFF
--- a/pex.go
+++ b/pex.go
@@ -28,16 +28,10 @@ type pexEvent struct {
 	f    pp.PexPeerFlags
 }
 
-// Combines the node addr, as required for pp.PexMsg.
-type pexMsgAdded struct {
-	krpc.NodeAddr
-	pp.PexPeerFlags
-}
-
-// Makes generating a PexMsg more efficient.
+// facilitates efficient de-duplication while generating PEX messages
 type pexMsgFactory struct {
-	added   map[string]pexMsgAdded
-	dropped map[string]krpc.NodeAddr
+	added   map[addrKey]pexEvent
+	dropped map[addrKey]pexEvent
 }
 
 func (me *pexMsgFactory) DeltaLen() int {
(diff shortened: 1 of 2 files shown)


yarikk:
@anacrolix here's the optimisation I had in mind – comments are welcome!

pex.go
```diff
@@ -28,16 +29,10 @@ type pexEvent struct {
...
-	krpc.NodeAddr
-	pp.PexPeerFlags
-}
-
-// Makes generating a PexMsg more efficient.
+// facilitates efficient de-duplication while generating PEX messages
 type pexMsgFactory struct {
-	added   map[string]pexMsgAdded
-	dropped map[string]krpc.NodeAddr
+	added   map[addrKey]pexEvent
```
anacrolix, line 34: I don't think I like having the pexEvent as values in `added` and `dropped`. By the time they're stored there, we've already processed the event type, so it's redundant data. I'm left wondering why it's there.

anacrolix: It's even worse for dropped events, since the flags don't mean anything either there.

pex.go
```diff
@@ -46,90 +41,78 @@ func (me *pexMsgFactory) DeltaLen() int {
...
 func (me *pexMsgFactory) PexMsg() (ret pp.PexMsg) {
 	for key, added := range me.added {
-		switch len(key) {
-		case compactIpv4NodeAddrElemSize:
-			ret.Added = append(ret.Added, added.NodeAddr)
-			ret.AddedFlags = append(ret.AddedFlags, added.PexPeerFlags)
-		case compactIpv6NodeAddrElemSize:
-			ret.Added6 = append(ret.Added6, added.NodeAddr)
-			ret.Added6Flags = append(ret.Added6Flags, added.PexPeerFlags)
+		addr, ok := nodeAddr(added.addr)
```
anacrolix, line 92: I think I also prefer this the way it was: We'd already done the separation of IPv6 and IPv4 at the key level, here we're doing it again, and lazily generating the node addr. We need to know that addrs we've counted against hte pex msg delta len actually count. If we fail to add a node addr later, our calculations were out.

pex.go
```diff
@@ -138,14 +121,6 @@ func (me *pexMsgFactory) PexMsg() (ret pp.PexMsg) {
 	return
 }
 
-func mustNodeAddr(addr net.Addr) krpc.NodeAddr {
```
anacrolix, line 141: Moving this is okay, but it might actually be relevant in non-test code and end up coming back.

pex_test.go
```diff
@@ -288,61 +303,98 @@ func TestPexGenmsg(t *testing.T) {
...
+		c := addrgen(npeers)
+		for addr := range c {
+			s.Add(&PeerConn{peer: peer{RemoteAddr: addr}})
+			s.Genmsg(0)
+		}
 	}
-	f := pp.PexPrefersEncryption | pp.PexOutgoingConn
+}
+
+// obtain at least 5 points, e.g. to plot a graph
```
anacrolix, line 349: 👍🏻

anacrolix:
Thanks for the work. The benchmarks and tests are looking fantastic. A few things I'm not 100% on.

pex.go
yarikk: I can see your concern: there's an overhead in storing those entries wrapped as an event. However, there's a win, too: the resulting code is cleaner, with much reduced exposure to the `krpc.*` types which reflect how the data is represented on the wire and are clumsy to deal with outside of the serialisation primitives. Consider this an investment into simplicity. It will pay back, eventually.

pex.go
yarikk: Again, the overhead of checking for v4/v6 is little. The cost of optimising is an exposure to the implementation details of what goes into the key and how. With a yield that low, it just not worth the added complexity.

pex.go
yarikk: Please, no, the library has too much panics wired in already. See, if we skip an address which, somehow, ended up broken (haven't seen any by far), it's fine. No need to panic.

yarikk:
@anacrolix Thanks for your feedback, much appreciated. I responded under each concern raised with explanations why I chose to go one way and not the other – I hope it clears things up. Also, good call adding maps to speed up the de-duplications – nice job!

anacrolix:
Okay, let's give it a spin.

PR #438, opened 2020-11-12
Optimise PEX by avoiding intermediate storage

yarikk:
Turns out, storing the PEX message entries in maps for later extraction at message generation stage is very expensive. If we prepare the lists without storing in maps we gain a huge win for the initial message. Here are the new benchmarks:

```
BenchmarkPexInitial4-8     	  217302	      5089 ns/op
BenchmarkPexInitial50-8    	   24865	     48012 ns/op
BenchmarkPexInitial100-8   	   11413	     90340 ns/op
BenchmarkPexInitial200-8   	    6747	    189316 ns/op
BenchmarkPexInitial400-8   	    2803	    438807 ns/op
```

Comparing these against the "after" benchmark of #435, we may observe significant improvement:

Screenshot 2020-11-12 at 21 38 26

Accounting for reconnections may become slower but those, at later stages, come on top of the log processed earlier, amortising the cost.

### DIFF
--- a/pex.go
+++ b/pex.go
@@ -31,8 +31,9 @@ type pexEvent struct {
 
 // facilitates efficient de-duplication while generating PEX messages
 type pexMsgFactory struct {
-	added   map[addrKey]pexEvent
-	dropped map[addrKey]pexEvent
+	msg     pp.PexMsg
+	added   map[addrKey]struct{}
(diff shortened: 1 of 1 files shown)


anacrolix:
image

PR #440, opened 2020-11-18
Readme: downstream projects info.

pataquets:
No longer using anacrolix/torrent but libtorren﻿t. However, I didn't want to delete, because it has been using it for quite a while.


pataquets:
cc: @jpillora @boypt FYI

Issue #342, opened 2019-11-09
torrent-create always sets comment 'yoloham'


rwenz3l:
Based on the defaults defined here: 
https://github.com/anacrolix/torrent/blob/dfcb8524874306491949934638d05161329014a3/metainfo/metainfo.go#L60-L65

all torrents created with `torrent-create` include a comment.

Tested with:
```
go get github.com/anacrolix/torrent/cmd/...
dd if=/dev/random of=~/test.mp4 bs=1m count=100
torrent-create ~/test.mp4 > ~/test.mp4.torrent
```

I consider this a bug, as I do not expect a comment when there is none specified, especially not something like 'yoloham'.

anacrolix:
@rwenz3l a PR would be welcomed. It's probably enough just to remove the "yoloham" comment assignment.

anacrolix:
@rwenz3l are you still interested in this?

rwenz3l:
hey @anacrolix, I had this in my backlog for quite some time. I could def. take care of this, but not sure when I will actually get around doing it. 

Am I interested: Yes. Am I going to work on it anytime soon? Maybe.

PR #443, opened 2020-12-09
Update metainfo.go

Aranjedeath:
delete "yoloham" so that is no longer the default comment string on torrents made without a comment specified. Should resolve https://github.com/anacrolix/torrent/issues/342


anacrolix:
😿

PR #446, opened 2020-12-14
PEX: impede full-meching in tracker-less swarms by adding a cooldown minute

yarikk:
Tracker-less PEX-only swarms quickly turn fully meshed (every peer has a connection to every other peer), contrary to tracker-backed swarms which have periodic poll interval for contacting trackers.  Proposed  is to mimic the throttling of peering info supply by adding a  minute-long cool down period (which  coincides with  the shortest  tracker announce interval) during which incoming PEX messages are discarded without processing.

Thoughts and comments are welcome, @anacrolix

### DIFF
--- a/pexconn.go
+++ b/pexconn.go
@@ -88,6 +88,15 @@ func (s *pexConnState) Share(postfn messageWriter) bool {
 
 // Recv is called from the reader goroutine
 func (s *pexConnState) Recv(payload []byte) error {
+	if !s.torrent.wantPeers() {
+		s.dbg.Printf("peer reserve ok, incoming PEX discarded")
+		return nil
+	}
+	if time.Now().Before(s.torrent.pex.rest) {
+		s.dbg.Printf("in cooldown period, incoming PEX discarded")
+		return nil
+	}
+
 	rx, err := pp.LoadPexMsg(payload)
 	if err != nil {
 		return fmt.Errorf("error unmarshalling PEX message: %s", err)
@@ -100,8 +109,10 @@ func (s *pexConnState) Recv(payload []byte) error {
 	peers.AppendFromPex(rx.Added6, rx.Added6Flags)
 	peers.AppendFromPex(rx.Added, rx.AddedFlags)
 	s.dbg.Printf("adding %d peers from PEX", len(peers))
-	s.torrent.addPeers(peers)
-	// s.dbg.Print("known swarm now:", s.torrent.KnownSwarm())
+	if len(peers) > 0 {
+		s.torrent.pex.rest = time.Now().Add(pexInterval)
+		s.torrent.addPeers(peers)
+	}
 
 	// one day we may also want to:
 	// - check if the peer is not flooding us with PEX updates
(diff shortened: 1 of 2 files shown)


anacrolix:
Are you seeing performance issues in fully-meshed scenarios in your usage? If so, is that cost arising out of calls into Torrent.addPeers?

pexconn.go
```diff
@@ -100,8 +109,10 @@ func (s *pexConnState) Recv(payload []byte) error {
 	peers.AppendFromPex(rx.Added6, rx.Added6Flags)
 	peers.AppendFromPex(rx.Added, rx.AddedFlags)
 	s.dbg.Printf("adding %d peers from PEX", len(peers))
-	s.torrent.addPeers(peers)
-	// s.dbg.Print("known swarm now:", s.torrent.KnownSwarm())
+	if len(peers) > 0 {
+		s.torrent.pex.rest = time.Now().Add(pexInterval)
+		s.torrent.addPeers(peers)
```
anacrolix, line 114: Would prefer not to guard this call, unless there's some special treatment of 0 peers in addPeers.

pexconn.go
```diff
@@ -88,6 +88,15 @@ func (s *pexConnState) Share(postfn messageWriter) bool {
 
 // Recv is called from the reader goroutine
 func (s *pexConnState) Recv(payload []byte) error {
+	if !s.torrent.wantPeers() {
```
anacrolix, line 91: This is tricky: The peer high and low water levels are designed for this, and addPeers should take care of that. Per my other comment, are we avoiding addPeers for performance costs?

pexconn.go
yarikk: This guard is mainly to avoid rewinding the cool down timer upon receipt of the PEX messages which carry drops only.

pexconn.go
yarikk: Indeed, and `wantPeer()` is one the parts of the high/low-watermak mechanism. Without it, PEX keeps feeding peers regardless whether the low watermark is reached; `addPeers` only cuts off at high watermark.

yarikk:
> Are you seeing performance issues in fully-meshed scenarios in your usage? If so, is that cost arising out of calls into Torrent.addPeers?

Full meshing puts more pressure on seeders, hurting the total throughput. The proposed change does not completely prevent it but makes the building up more gradual.

anacrolix:
> P.S. What's does the "fix cache" label imply?

I have no idea 😆 .

anacrolix:
@yarikk I'm happy to merge this, if you've been using it successfully in your own use-cases.

yarikk:
> @yarikk I'm happy to merge this, if you've been using it successfully in your own use-cases.

Yes, indeed. Although it doesn't stop full meshing completely, it certainly helps to pace it – enough to warrant keeping this patch in. Thanks for considering this!

yarikk:
Example of paced full-meshing: 
136342305_885042265598936_7463551091756080065_n

Issue #416, opened 2020-06-23
Torrent not available after a while when using WebTorrent


2m:
I used `torrent download -seed 'magnet:?xt=urn:btih:...'` to download and then seed the torrent via WebTorrent.

If I try to download the torrent (using instant.io or btorrent.xyz) just after download completes, so after

```
downloading "...": 288 MB/288 MB, 2197/2197 pieces completed (0 partial)
```

is printed to the logs, then everything is alright.

But then, after waiting for a while I see the following in the logs:

```
2020-06-23T11:19:49+0300 INFO  tracker_client.go:100: tracker client for "wss://tracker.btorrent.xyz": websocket instance ended: read message error: websocket: close 1006 (abnormal closure): unexpected EOF
2020-06-23T11:20:32+0300 INFO  tracker_client.go:100: tracker client for "wss://tracker.openwebtorrent.com": websocket instance ended: read message error: websocket: close 1006 (abnormal closure): unexpected EOF
```

And immediately after those logs, I am unable to download the seeded file via WebTorrent using the websites mentioned above.

anacrolix:
> And immediately after those logs, I am unable to download the seeded file via WebTorrent using the websites mentioned above.

Do you mean that other BitTorrent clients using the trackers you mentioned, aren't able to locate your local client that should be seeding now? Are you using your own torrent, for which you expect only your own client to be seeding?

There's some timing behaviour with the Webtorrent trackers that seem to expect announces to occur every minute, or the websocket connection is closed. The anacrolix/torrent implementation currently doesn't re-announce unless a submitted offer is taken up. However it will reconnect the websocket if it is disconnected. Are you seeing the websocket disconnect errors on a persistent basis after you finish the download?

2m:
> Do you mean that other BitTorrent clients using the trackers you mentioned, aren't able to locate your local client that should be seeding now?

Yes. That is correct. Immediately after I finish downloading and start seeding the torrent with `anacrolix/torrent` I can reach the same torrent via instant.io or btorrent.xyz. But after a while, after I see the mentioned log messages, the torrent is not available via instant.io or btorrent.xyz even if `anacrolix/torrent` is still running in the seed mode.

> Are you using your own torrent, for which you expect only your own client to be seeding?

Not exactly my own, but with very little seeders. And most likely without any WebTorrent seeders. So when I download that torrent using `anacrolix/torrent` I think it gets downloaded from regular seeds, and then I become the only WebTorrent seeder.

[...]

PR #455, opened 2021-01-05
Ping websocket to ensure connection remains online.

DigitalAlchemist:
OpenWebTorrent disconnects clients after two minutes of inactivity. In a scenario where an
announce occurs and no other activity occurs for the (default) 120-second idle timer, the
websocket will be torn down by OpenWebTorrent. This in turn causes the tracker client to wait one
minute, reconnect, and reannounce its peers.

This patch sends a websocket.PingMessage every 60 seconds to maintain the connection.

Relates to #416


DigitalAlchemist:
Thanks @anacrolix !

anacrolix:
Thanks for the contributions.

PR #482, opened 2021-05-07
Change ClientImpl to ClientImplCloser

gitpubber:
We cannot close underlying Boltdb PieceCompletion if it is not closable.


anacrolix:
Do you know if it would be possible to return the `*fileClientImpl` all the way through?

PR #491, opened 2021-05-12
DoOnClose() function and Build Tags

gitpubber:
Hello @anacrolix ,

1) I added DoOnClose() function which accepts function and appends it to Client's onClose array. I needed to append function to Client's onClose and now I can do it.

2) so, I analyzed my binary with goweight and realized that crawshaw.io/sqlite3 contributed to 13 MB of binary size. Simply it is optional dependency and we should be able to not build them if necessary . 

I added 3 build tags. `noboltdb` tag doesn't build boltdb piececompletion. `nosqlite3` doesn't build crawshaw.io/sqlite3 . These tags are don't affect normal workflow and only helps when you want to disable them.

### DIFF
--- a/client.go
+++ b/client.go
@@ -116,6 +116,13 @@ func writeDhtServerStatus(w io.Writer, s DhtServer) {
 	spew.Fdump(w, dhtStats)
 }
 
+// Passed Function will be executed when the Client Closes
+func (cl *Client) DoOnClose(f func()) {
+	cl.rLock()
+	defer cl.rUnlock()
+	cl.onClose = append(cl.onClose, f)
+}
+
 // Writes out a human readable status of the client, such as for writing to a
 // HTTP status page.
 func (cl *Client) WriteStatus(_w io.Writer) {
--- a/storage/boltPieceCompletion.go
+++ b/storage/boltPieceCompletion.go
@@ -1,3 +1,5 @@
+// +build !noboltdb
+
 package storage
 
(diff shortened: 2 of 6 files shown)


anacrolix:
Looks good, but why can't you register your own Close handlers elsewhere? Is it possible to manage those outside of Client? Just curious what your use case is.

gitpubber:
> Looks good, but why can't you register your own Close handlers elsewhere? Is it possible to manage those outside of Client? Just curious what your use case is.

Currently, you can't close PieceCompletion  on Client Close while  NewClient function (which takes ClientConfig as arg) closes it.

See https://github.com/anacrolix/torrent/blob/b8e1c9539907b177a7dac66c95304c0dbe91475f/client.go#L214

How can we do this for custom piece completion?

anacrolix:
If you have your own piece completion, you should be initializing it outside the Client, and closing it after you close the Client. If this isn't possible or reflected in the API, it should be fixed (submit an issue or point me to where its violating this).

gitpubber:
@anacrolix Can you merge Build Tags atleast . It helps to really slim down Binaries :)

anacrolix:
Yes. Sorry the PR needs to be updated.

gitpubber:
> Yes. Sorry the PR needs to be updated.

ah, i didn't see. but, boltdb only builds on cgo.

guess my issue is solved

PR #496, opened 2021-05-17
Added Methods to Update Rate Limiters and Change IPBlocklist

gitpubber:
@anacrolix , This is necessary to change those values without restarting Torrent Client

### DIFF
--- a/client.go
+++ b/client.go
@@ -160,6 +160,27 @@ func (cl *Client) WriteStatus(_w io.Writer) {
 	}
 }
 
+// Change DownloadRateLimiter of Client
+func (cl *Client) ChangeDownloadRateLimiter(r *rate.Limiter) {
+	cl.lock()
+	defer cl.unlock()
+	cl.config.DownloadRateLimiter = r
+}
+
+// Change UploadRateLimiter of Client
+func (cl *Client) ChangeUploadRateLimiter(r *rate.Limiter) {
+	cl.lock()
+	defer cl.unlock()
+	cl.config.UploadRateLimiter = r
+}
+
+// Change IPBlocklist of Client
+func (cl *Client) ChangeIPBlockList(bl iplist.Ranger) {
+	cl.lock()
+	defer cl.unlock()
+	cl.ipBlockList = bl
+}
+
 // Filters things that are less than warning from UPnP discovery.
 func upnpDiscoverLogFilter(m log.Msg) bool {
 	level, ok := m.GetLevel()


anacrolix:
It's a fair PR, but I'm certain that iplist.Ranger and *rate.Limiter can be altered out of band with the Client. It doesn't need to be aware, as long as synchronization is present on the implementations chosen. I'm sure that *rate.Limiter does, see https://pkg.go.dev/golang.org/x/time/rate. A thread-safe version of iplist.Ranger will fix the other issue.

I really don't want to grow the API more than it needs to, it already has a lot of legacy cruft.

gitpubber:
@anacrolix ,i will directly manipulate *rate.Limiter then. what can be done about iplist.Ranger if i need to change it(i just wanted to update blocklist once per day)

anacrolix:
@gitpubber which iplist.Ranger implementation are you using? You could make a PR to add locking to the implementation, and some methods to allow updating the list underneath. I will evaluate how difficult it is based on which implementation being used, and possibly accept the method to change the IP blocklist in this PR if it looks too difficult or not possible.

gitpubber:
@anacrolix , i am using github.com/anacrolix/torrent/iplist itself.


```
func FetchBlockList() (*iplist.IPList, error) {
	resp, err := http.Get(BlockListURL)
	if err != nil {
		return nil, err
	}

	// Limit Response
	blocklistReader := io.LimitReader(resp.Body, 20971520)

	// Extract file
	gzipReader, err := gzip.NewReader(blocklistReader)

	if err != nil {
		log.Printf("Error extracting blocklist: %s", err)
		return nil, err
	}

	// Read blocklist
	blocklist, err := iplist.NewFromReader(gzipReader)
	if err != nil {
	        log.Println("Error reading blocklist: ", err)
		return nil, err
	}
	_ = resp.Body.Close()
	log.Println("Loading blocklist of ", blocklist.NumRanges(), " Ranges")
	return blocklist, nil
}
```

gitpubber:
and i would like to update this blocklist once per day. what do you suggest @anacrolix  ?

anacrolix:
Nice. I'd create something like
```
type SynchronizedRanger struct {
    mu sync.RWMutex
    inner iplist.Ranger
}
```
With Lookup and NumRanges that make use of the inner Ranger, using a read lock. Then in your daily blocklist fetch, I'd build a new Ranger, lock the SynchronizedRanger and replace the inner.

gitpubber:
Thank you @anacrolix , i will do as you said

PR #499, opened 2021-05-21
Added build tags to disable few packages if necessary

gitpubber:
This adds 2 buildtags. `noboltdb` doesn't build boltdb and `nosqlite` doesn't build sqlite3 . This helps in decreasing size of binaries.

### DIFF
--- a/storage/bolt-piece-completion.go
+++ b/storage/bolt-piece-completion.go
@@ -1,3 +1,6 @@
+//go:build !noboltdb
+// +build !noboltdb
+
 package storage
 
 import (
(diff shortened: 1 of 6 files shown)


yarikk (approved):
Thanks, this may come handy!

yarikk:
Alternatively, could we make it opt-in, and have `!boltdb` and `!sqlite` as build tags. Thoughts, @anacrolix?

gitpubber:
@yarikk i think we better have noboltdb and nosqlite . currently anacrolix/torrent automatically selects boltdb/sqlite based on whether cgo is not present/present.

if we want buildtags to include those, then it becomes hassle for downstream

anacrolix:
Thanks. I tried something like this, and the unit tests will fail though. Is that something you could include in the patch? No worries if it doesn't interest you.

anacrolix:
Perfectly good PR, I don't want to hold it up.

Issue #500, opened 2021-05-21
Use gobwas/ws instead of gorilla/websocket


gitpubber:
This package uses gorilla/websocket for webtorrent which many criticise as being slow and large.few other packages of websocket like gobwas/ws which are more lightweight and fast are also present.

Why not use them? why use gorilla/websocket?

https://www.freecodecamp.org/news/million-websockets-and-go-cc58418460bb/

https://github.com/gobwas/ws

it could be better to use these packages, no?

gitpubber:
@anacrolix , i saw the codebase which uses gorilla/websocket . i think it's not large(being single file).

can you evaluate https://github.com/gobwas/ws to replace gorilla/websocket ?

gitpubber:
https://github.com/anacrolix/torrent/blob/master/webtorrent/tracker_client.go

gitpubber:
I could have tried to rewrite this, but i don't have much knowledge :(

please help @anacrolix . i would like to use gobwas/ws in my app and i would not like to have 2 websocket dependencies

anacrolix:
Thanks for the suggestion. It just comes down to gorilla/websocket being the most popular after moving on from an golang/x websocket implementation that didn't cope with certain types of usage in some of my projects. I have no attachment to gorilla/websocket, but also no time to work on this unless it affects me currently.

gitpubber:
@anacrolix Done.

see https://github.com/anacrolix/torrent/pull/503

PR #503, opened 2021-05-25
Use gobwas/ws instead of gorilla/webtorrent

gitpubber:
Closes https://github.com/anacrolix/torrent/issues/500

This successfully compiles and is tried on my computer.

With this patch, better gobwas/ws will be used instead of gorilla/webtorrent


anacrolix:
I think the *net.Conn should be just net.Conn, otherwise looks good.

gitpubber:
> I think the *net.Conn should be just net.Conn, otherwise looks good.

You have checked whether *websocket.Conn is nil .What should be the logic there?

gitpubber:
@anacrolix can you please tell what should be done here?

https://github.com/anacrolix/torrent/blob/3e248a7fcad4662eda79e8dfbfbdfd6f30f6e90e/webtorrent/tracker_client.go#L238

anacrolix:
You can still compare net.Conn with nil. It will work fine.

anacrolix:
In the future you can just push more commits to the PR, or change your commits and force push. No need to open another PR.

PR #518, opened 2021-06-07
Allow Storage Backends to do their own Hashing

Zilog8:
- Describes an optional interface 'SelfHashing' that a storage backend's type implementing 'PieceImpl' may also implement in order to calculate piece hashsums itself.

- Alters the 'hashPiece' function in the torrent package to look for types implementing 'SelfHashing' . If not implemented, calculate the hash as usual.


anacrolix:
Looks very clean, just waiting on CI. Did this work for your purposes?

Zilog8:
It's working well with my proof-of-concept backend. The backend on the other hand is still unfit for public consumption :sweat_smile:

anacrolix:
Thanks!

PR #521, opened 2021-06-17
Updated roaring

xackery:
Prior to above, you get the error:
```
go get: github.com/willf/bitset@v1.1.11 updating to
        github.com/willf/bitset@v1.2.0: parsing go.mod:
        module declares its path as: github.com/bits-and-blooms/bitset
                but was required as: github.com/willf/bitset
```
This is an indirect dependency of roaring - github.com/RoaringBitmap/roaring v0.7.2 // indirect

It appers willf changed to bits-and-blooms and broke older versions.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/anacrolix/torrent
 require (
 	bazil.org/fuse v0.0.0-20200407214033-5883e5a4b512
 	crawshaw.io/sqlite v0.3.3-0.20210127221821-98b1f83c5508
-	github.com/RoaringBitmap/roaring v0.6.0 // indirect
+	github.com/RoaringBitmap/roaring v0.7.2 // indirect
 	github.com/alexflint/go-arg v1.3.0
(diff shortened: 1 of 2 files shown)


anacrolix:
How can I reproduce this?

xackery:
This was part of my attempts to debug that 'e' issue, i don't think it's worth PRing since there's no other benefit to updating Roaring to 0.72, but just a FYI you have a indirect ref from it to bitset and it changed org name, for now closing.

anacrolix:
Thanks for the followup.

anacrolix:
@xackery I came across this problem too, but there seems to be another issue deeper inside willf's repos.

gitpubber:
@anacrolix  why can't you update roaringbitmaps to latest version. Won't it solve the issue ?

anacrolix:
I had no luck trying to solve this. Please share if you find a solution. The goal is to run go get -u with no errors, and a successful build.

Issue #479, opened 2021-05-04
Stop Seeding during runtime


gitpubber:
How can I do it? There is ClientConfig.Seed but how can I change it during runtime ?

anacrolix:
This will need to be exposed. It probably wouldn't hurt to expose stuff like download/upload ratio too to enable users to make use of that in configuring seeding.

gitpubber:
Can you please expose seed and download/upload ratio ?

anacrolix:
I will look into it if I have time, cheers.

gitpubber:
Hello, what's the progress on this? also, to change iplist.Ranger , should we lock client's mutex before changing? if yes, can you please create wrapper function to perform so?

anacrolix:
Sorry, I'm not working on this currently.

anacrolix:
This remains a valid enhancement. Possibly related to https://github.com/anacrolix/torrent/issues/155.

gitpubber:
@anacrolix ,
Possible workarounds include ranging through torrents in client and if they are completed, disable data upload

anacrolix:
Yeah, there are definitely workarounds. Good thinking.

gitpubber:
https://github.com/anacrolix/torrent/blob/b21aebeaae6a308d97e9f952c78639bc594e73fd/torrent.go#L1363

So, Config is directly been referenced. so,it would be easy to implement this feature,no?

gitpubber:
@anacrolix any update on this?

gitpubber:
@anacrolix can you provide method to change Seed value of config(while mutex locking it) ?

It must solve this issue.right?

anacrolix:
Yes that would work. There might be a few condition values to tickle or something too.

gitpubber:
https://github.com/anacrolix/torrent/pull/527 @anacrolix

PR #527, opened 2021-06-30
Enable/Disable Seeding during Runtime

gitpubber:
Closes https://github.com/anacrolix/torrent/issues/479

### DIFF
--- a/torrent.go
+++ b/torrent.go
@@ -1444,7 +1444,7 @@ func (t *Torrent) seeding() bool {
 	if cl.config.NoUpload {
 		return false
 	}
-	if !cl.config.Seed {
+	if !cl.IsSeeding() {
 		return false
 	}
 	if cl.config.DisableAggressiveUpload && t.needData() {
--- a/client.go
+++ b/client.go
@@ -92,6 +92,23 @@ func (cl *Client) BadPeerIPs() []string {
 	return cl.badPeerIPsLocked()
 }
 
+func (cl *Client) IsSeeding() (ret bool) {
+	cl.rLock()
+	defer cl.rUnlock()
+	if cl.config != nil {
+		return cl.config.Seed
+	}
+	return
+}
+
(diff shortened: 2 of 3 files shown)


gitpubber:
@anacrolix

anacrolix:
Would you like to add some tests? Perhaps a test that it works to switch seeding on after initially starting off.

gitpubber:
@anacrolix , i have written tests

gitpubber:
@anacrolix  , I have also made checking whether it is seeding , mutex protected . I believe there are no dataraces here and code is fine.

torrent.go
```diff
@@ -1444,7 +1444,7 @@ func (t *Torrent) seeding() bool {
 	if cl.config.NoUpload {
 		return false
 	}
-	if !cl.config.Seed {
+	if !cl.IsSeeding() {
```
anacrolix, line 1447: I don't know if it's okay to call a method here that obtains a read lock on the Client.

gitpubber: Why not ? Does it acquire recursive read lock here ? 

Oh, Can we have seperate mutex lock for Seeding ?

gitpubber:
@anacrolix  I think it must be correct now.

torrent.go
gitpubber: @anacrolix  I think next commit solves it

anacrolix: I don't think a separate mutex is the requirement, it's recursively locking it.

anacrolix:
I'm not sure I want to merge this yet, I haven't got enough testing, and I'm not sure how I want to expose this.

torrent.go
gitpubber: @anacrolix oh,ok. what might be optimal solution for this?

gitpubber:
@anacrolix No problem. this means that Seed can't be changed during runtime. I wish that you come up with some solution so that Seed variable can be changed during runtime

PR #529, opened 2021-07-15
call decrementPendingWrites() first to free storageLock, before obtai…

fangjun1024:
It will get a deadlock happened if torrent.Drop() and torrent.pieceHasher() execute at the same time. 

torrent.Drop() is holding _Client.Lock_ and want to obtain _Torrent.storageLock_, but torrent.pieceHasher() already take _Torrent.storageLock_ and wait to check pending writes by waitNoPendingWrites().  

On the other side, incrementPendingWrites() is called in receiveChunk(),  but decrementPendingWrites() will not be called before get _Client.Lock_ , which is held by torrent.Drop(). 

So we could call decrementPendingWrites() first before obtain Client.Lock, and it should be called anyway if writeChunk() is done.

### DIFF
--- a/peerconn.go
+++ b/peerconn.go
@@ -1324,6 +1324,7 @@ func (c *Peer) receiveChunk(msg *pp.Message) error {
 		defer cl.lock()
 		concurrentChunkWrites.Add(1)
 		defer concurrentChunkWrites.Add(-1)
+		defer piece.decrementPendingWrites()
 		// Write the chunk out. Note that the upper bound on chunk writing concurrency will be the
 		// number of connections. We write inline with receiving the chunk (with this lock dance),
 		// because we want to handle errors synchronously and I haven't thought of a nice way to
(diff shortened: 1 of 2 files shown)


anacrolix:
There's a data race reported in the tests.

anacrolix:
That's a very convoluted race, but it checks out in my static analysis. Good catch!

anacrolix:
The data race occurs due to the read here: https://github.com/anacrolix/torrent/blob/162745651660ca7c3a58546960737112672d8e64/peerconn.go#L1347. I wonder if that can be fixed somehow. It might be possible to instead fix the fact that pieceHasher holds the storage lock while waiting on the pending writes, or do the storage closing in Torrent.Drop asynchronously.

anacrolix:
I think the client lock should be held when decrementing the writes, as that line needs to make a synchronous decision about queuing a piece check. What about closing the storage asynchronously from Torrent.Drop instead?

fangjun1024:
sorry not noticed the Data Race error, we cloud protect pendingWrites with _pendingWritesMutex_. 

PR is updated, please take a look.

anacrolix:
Thanks, you've fixed it the same way I first tried. I'm not sure that it works, some more testing is required. In the meanwhile I've included the asynchronous storage closing I mentioned as an alternative, on master.

anacrolix:
Thank you for the very tricky find. I believe this is fixed, with some more fixes coming to deal with the fact that close now does stuff asynchronously. Let me know if you find anything else.

PR #535, opened 2021-07-26
Change QueryEscape to PathEscape

YenForYang:


anacrolix:
I think this was intentionally QueryEscape for compatibility with S3. If that seems wrong, we should double check the git logs.

YenForYang:
Yup I made a mistake. This doesn't appear to be the fix for the url-encoding issue. I'll close this.

PR #536, opened 2021-07-26
Eliminate redundant copies of DhtNodes

YenForYang:
Safe use of `unsafe` to directly convert `[]Node` to `[]string`

### DIFF
--- a/spec.go
+++ b/spec.go
@@ -3,6 +3,7 @@ package torrent
 import (
 	"github.com/anacrolix/torrent/metainfo"
 	"github.com/anacrolix/torrent/storage"
+	"unsafe"
 )
 
 // Specifies a new torrent for adding to a client. There are helpers for magnet URIs and torrent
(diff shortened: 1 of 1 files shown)


anacrolix:
Performance reasons or just because you can? 🤣

YenForYang:
I honestly think it's a worthwhile use of `unsafe`. I think it's more readable, but it also happens to reduce code size and eliminate an extra allocation + copy. 

All that aside, I do think there is a mistake in the current code (which ironically would have been avoided with the use of `unsafe`):

```
ret = make([]string, len(mi.Nodes))
for _, node := range mi.Nodes {
	ret = append(ret, string(node))
}
```

Should probably be something like
```
ret = make([]string, len(mi.Nodes))
for i := range mi.Nodes {
	ret[i] = string(mi.Nodes[i])
}
```

But I'm not sure if the empty nodes are intentional.

anacrolix:
Well that bit about allocating empty strings is definitely an error. I'll fix that now.

anacrolix:
@YenForYang are you still interested in getting this one in?

YenForYang:
Of course! You can say that I get pretty obsessive over optimizing in Go, but it's nice to find an `unsafe` tweak that isn't too ugly. We should use it :D

anacrolix:
I was able to crash things easily if I made changes to `type Node` with this change. Covariance/contravariance whichever one it is for instance would be nice.

YenForYang:
Not sure what you mean exactly. Do you mean change as in `type Node string` to `type Node SomeOtherType`? In that case would compile time checks (i.e. emulate `_Static_assert`/`static_assert`) fix this issue?

I mean, if you change `Node` to be defined by `SomeOtherType` I'd assume you would have to change the closure anyway.

PR #537, opened 2021-07-26
Fix info_hash url encoding

YenForYang:


YenForYang:
Ah after some testing it looks like even `PathEscape` sometimes uses `+` instead of `%20`. I think resorting to `strings.Replace()` may be the only way

anacrolix:
I'm confused, doesn't QueryEscape already encode spaces to `%20`? You're welcome to add the logging change if it helps.

anacrolix:
@YenForYang any update on this?

YenForYang:
Doesn't look like QueryEscape uses %20 consistently. I've been doing an explicit strings.Replace and it works great.

YenForYang:
I'll submit my current changes, but I doubt you'll like the looks of it. I basically ditched using `url.Values` and decided to build the string from scratch. I actually find it more readable and it hides less "magic" IMO. I think finer control over building the query is ultimately needed; the url package simply doesn't provide this.

YenForYang:
Just a side note -- We probably won't get more control of the percent encoding with QueryEscape, given this issue: https://github.com/golang/go/issues/13982

anacrolix:
@YenForYang what about https://github.com/anacrolix/torrent/pull/591 instead?

PR #542, opened 2021-08-09
Add linter CI

boramalper:
Signed-off-by: Bora M. Alper <bora@boramalper.org>


anacrolix:
It's a nice idea. Many of the commands are failing for weird reasons. The security check has a lot of false positives.

boramalper:
@anacrolix Made linters skip some checks to remove the noise, and fixed many warnings except a few which I believe are better handled by you. Let me know if there is anything else you would want to see.

anacrolix:
This looks more manageable. Thanks!

anacrolix:
The linter has started reporting a lot more stuff. I had a tidy up, but I still get this output: https://github.com/anacrolix/torrent/runs/3368658878. There's no line numbers, it doesn't say what Go version. I don't know if this is the workflow or golangci-lint that aren't being helpful.

boramalper:
Maybe it's better to call gofmt, staticcheck, and go vet explicitly? I agree that there is too much "magic" at the moment and it is not clear what checks are being run and where do the errors stem. Would you like another PR?

anacrolix:
If you like. I'm not keeping up with the world of linters and GitHub Actions, but appreciate their value.

PR #544, opened 2021-08-10
cmd/torrent/main.go does not compile

bsergean:
(unreferenced AnnounceCmd)

```
$ go run cmd/torrent/main.go
# command-line-arguments
cmd/torrent/main.go:167:2: undefined: AnnounceCmd
```


anacrolix:
Try `go run ./cmd/torrent` instead.

bsergean:
Ah, it works indeed. I'm a go noob ... thanks for the tip !

PR #552, opened 2021-08-27
Trim allocations when returning FilePieceState

YenForYang:
Use of `append` in `File.State()` can be eliminated in favor of precalculating piece count, calling `make()` and assigning directly. Removes the need of keeping track of remaining bytes and making additional checks when looping over pieces.

### DIFF
--- a/file.go
+++ b/file.go
@@ -101,26 +101,32 @@ type FilePieceState struct {
 }
 
 // Returns the state of pieces in this file.
-func (f *File) State() (ret []FilePieceState) {
+func (f *File) State() []FilePieceState {
 	f.t.cl.rLock()
 	defer f.t.cl.rUnlock()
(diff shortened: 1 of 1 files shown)


anacrolix:
It's a very explicit way to do it. The lazy version might be allocate with the cap set, make(a, b, cap). Have you had performance overhead here?

anacrolix:
@YenForYang I've noticed the theme with preallocating instead of creating slices with sufficient capacity and appending. I'm very happy to accept these changes if you can just link me to something showing it's faster. Cheers

YenForYang:
Made another change -- should be inlineable now (assuming its cost isn't too high). Seems like all my requests are on inlining lol.

anacrolix:
My eyes are glazing over checking this. Unless there's a performance issue in this function, it's hard to see that it's worth changing.

PR #554, opened 2021-08-28
Remove unsafe in slice cast to array pointer

YenForYang:
Go 1.17 allows direct casting from slice to pointer to array.

### DIFF
--- a/torrent.go
+++ b/torrent.go
@@ -16,7 +16,6 @@ import (
 	"sync"
 	"text/tabwriter"
 	"time"
-	"unsafe"
 
 	"github.com/RoaringBitmap/roaring"
 	"github.com/anacrolix/dht/v2"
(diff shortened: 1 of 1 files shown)


anacrolix:
I wonder if this is optimal anyway. It's creating a pointer to the temporary copy of the hash, which should also be in a slice cast from []byte.

anacrolix:
I fixed a few CI issues.

YenForYang:
Yeah, after looking at the assembly I'd say it's not optimal -- you're right. There is overhead in how Go checks if the slice can be safely cast to the array pointer.

There is still unnecessary overhead though since Go insists on bounds checking `hash[0]`, which sucks.

The optimal solution is ugly, as it usually is with `unsafe`...

```
...
piece.hash = (*metainfo.Hash)(unsafe.Pointer((*reflect.SliceHeader)(unsafe.Pointer(&hash)).Data))
```

This doesn't look like something you'd be ok with though, lol. I'm completely fine with it nonetheless :D. I guess I'll close for now.

anacrolix:
No worries. I'm more than happy to do unsafe stuff and throw convention under the bus when it's necessary, but otherwise I'd prefer to keep it foolproof.

PR #562, opened 2021-09-03
useTorrentSources: fix http request resource leak

YenForYang:
Additional tweaks to context passing and looping

### DIFF
--- a/client.go
+++ b/client.go
@@ -1207,46 +1207,47 @@ func (t *Torrent) MergeSpec(spec *TorrentSpec) error {
 }
 
 func useTorrentSources(sources []string, t *Torrent) {
-	for _, s := range sources {
-		go func(s string) {
-			err := useTorrentSource(s, t)
-			if err != nil {
+	ctx := context.Background()
+	for i := 0; i < len(sources); i += 1 {
+		s := sources[i]
+		go func() {
+			if err := useTorrentSource(ctx, s, t); err != nil {
 				t.logger.WithDefaultLevel(log.Warning).Printf("using torrent source %q: %v", s, err)
 			} else {
 				t.logger.Printf("successfully used source %q", s)
 			}
-		}(s)
+		}()
 	}
 }
 
-func useTorrentSource(source string, t *Torrent) error {
-	req, err := http.NewRequest(http.MethodGet, source, nil)
-	if err != nil {
-		panic(err)
-	}
-	ctx, cancel := context.WithCancel(context.Background())
+func useTorrentSource(ctx context.Context, source string, t *Torrent) (err error) {
+	ctxWithCancel, cancel := context.WithCancel(ctx)
 	defer cancel()
 	go func() {
 		select {
 		case <-t.GotInfo():
 		case <-t.Closed():
-		case <-ctx.Done():
+		case <-ctxWithCancel.Done():
 		}
 		cancel()
 	}()
-	req = req.WithContext(ctx)
-	resp, err := http.DefaultClient.Do(req)
-	if err != nil {
-		return err
+	var req *http.Request
+	if req, err = http.NewRequestWithContext(ctxWithCancel, http.MethodGet, source, nil); err != nil {
+		panic(err)
 	}
-	mi, err := metainfo.Load(resp.Body)
-	if err != nil {
+	var resp *http.Response
+	if resp, err = http.DefaultClient.Do(req); err == nil {
+		var mi metainfo.MetaInfo
+		err = bencode.NewDecoder(resp.Body).Decode(&mi)
+		resp.Body.Close()
+		if err == nil {
+			return t.MergeSpec(TorrentSpecFromMetaInfo(&mi))
+		}
 		if ctx.Err() != nil {
 			return nil
 		}
-		return err
 	}
-	return t.MergeSpec(TorrentSpecFromMetaInfo(mi))
+	return
 }
 
 func (cl *Client) dropTorrent(infoHash metainfo.Hash) (err error) {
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -1207,46 +1207,47 @@ func (t *Torrent) MergeSpec(spec *TorrentSpec) error {
...
 }
 
-func useTorrentSource(source string, t *Torrent) error {
-	req, err := http.NewRequest(http.MethodGet, source, nil)
-	if err != nil {
-		panic(err)
-	}
-	ctx, cancel := context.WithCancel(context.Background())
+func useTorrentSource(ctx context.Context, source string, t *Torrent) (err error) {
+	ctxWithCancel, cancel := context.WithCancel(ctx)
```
anacrolix, line 1224: Why not just clobber `ctx` here?

client.go
```diff
@@ -1207,46 +1207,47 @@ func (t *Torrent) MergeSpec(spec *TorrentSpec) error {
...
+	var req *http.Request
+	if req, err = http.NewRequestWithContext(ctxWithCancel, http.MethodGet, source, nil); err != nil {
+		panic(err)
 	}
-	mi, err := metainfo.Load(resp.Body)
-	if err != nil {
+	var resp *http.Response
+	if resp, err = http.DefaultClient.Do(req); err == nil {
+		var mi metainfo.MetaInfo
+		err = bencode.NewDecoder(resp.Body).Decode(&mi)
```
anacrolix, line 1241: Why not use metainfo.Load here? I think it actually has different behaviour, (nevermind that it probably shouldn't exist).

client.go
```diff
@@ -1207,46 +1207,47 @@ func (t *Torrent) MergeSpec(spec *TorrentSpec) error {
...
-	if err != nil {
-		return err
+	var req *http.Request
+	if req, err = http.NewRequestWithContext(ctxWithCancel, http.MethodGet, source, nil); err != nil {
+		panic(err)
 	}
-	mi, err := metainfo.Load(resp.Body)
-	if err != nil {
+	var resp *http.Response
+	if resp, err = http.DefaultClient.Do(req); err == nil {
```
anacrolix, line 1239: It's silly but Go is easier to read if the errors return early whenver possible.

client.go
```diff
@@ -1207,46 +1207,47 @@ func (t *Torrent) MergeSpec(spec *TorrentSpec) error {
 }
 
 func useTorrentSources(sources []string, t *Torrent) {
-	for _, s := range sources {
-		go func(s string) {
-			err := useTorrentSource(s, t)
-			if err != nil {
+	ctx := context.Background()
```
anacrolix, line 1210: It makes sense to make context part of useTorrentSource's signature, except we don't actually make use of it. Why not bind it to the lifetime of the Torrent or something, so it is cancelled if the torrent closes?

client.go
YenForYang: Just a style thing for me, I guess it can be changed if desired. It just makes it clear that `ctxWithCancel` is a new context derived from `ctx`. But I can also see how `ctx` might be accidentally used instead of `ctxWithCancel`, so I'm fine with either way.

client.go
YenForYang: Tbh I just didn't like how `metainfo.Load` returned a pointer instead of a value (especially since it declares `var mi metainfo.MetaInfo` only to return `&mi` -- thus relying on escape analysis). I wanted to restrict the lifetime of `mi` as much as possible, without relying on the GC to collect it later from the heap.

But if you insist on using `metainfo.Load`, I guess we could do something like
```
var mi *metainfo.MetaInfo
mi, err = metainfo.Load(resp.Body)
```
and rely on escape analysis to determine how to clean up `mi`. IMO it's better to just scope `mi`.

client.go
YenForYang: I can't say I agree with Go being easier to read if errors return early whenever possible. However, I definitely think that errors *should* return as soon as possible (regardless of being easier to read or not). If I'm not mistaken I think the errors here do return as soon as possible.

client.go
YenForYang: Yeah that's definitely true; I just thought of it more as a style thing for future possible use cases of `useTorrentSource` (no idea what else it could be used for though). But, yeah, if you're sure that `useTorrentSource` will only ever be used by `useTorrentSources` then it would be best to remove the parameter altogether. It would be even better if we could restrict its visibility to only `useTorrentSources` actually.

The restricting visibility part would require moving `useTorrentSource` to a closure (since nested functions cant be done in Go), and definitely make things uglier. If you're OK with that, I'm all for it.

anacrolix: No I'm fine with this, maybe a TODO to wire it in later to the lifetime of Torrent.

client.go
anacrolix: I think the original style is fine:
```
req, err := ...
if err != nil { ... }
resp, err := ...
if err != nil { ... }
```
Go is already horribly verbose, but after years of fighting it I think it's easiest just to do be straightforward with it.

client.go
anacrolix: Yep, keep this the way you have, although note escape analysis will probably put it on the heap anyway.

anacrolix (changes requested):
What was the original leak, the http Response body not being closed?

YenForYang:
Yeah response body not being closed caused the leak.

YenForYang:
I think I made all the changes you requested -- I inverted the `err == nil`s and ended up deciding to shadow `ctx` instead of creating `ctxWithCancel` to prevent confusion.

anacrolix:
I think I saw some extra newlines floating around. You might want to use go fmt.

PR #563, opened 2021-09-03
Preallocate torrentsAsSlice()

YenForYang:
Using `append` would be a better option if `torrentsAsSlice()` accepted a slice argument. I'd prefer that, but it would require some API changes probably. `make+copy` has better optimizations than `append` since Go 1.15, so it's best to use that and not make `append` do guesswork. Also saves a little work for the GC since extra pointers would be allocated.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1290,15 +1290,19 @@ func (cl *Client) WaitAll() bool {
 }
 
 // Returns handles to all the torrents loaded in the Client.
-func (cl *Client) Torrents() []*Torrent {
+func (cl *Client) Torrents() (ts []*Torrent) {
 	cl.lock()
-	defer cl.unlock()
-	return cl.torrentsAsSlice()
+	ts = cl.torrentsAsSlice()
+	cl.unlock()
+	return
 }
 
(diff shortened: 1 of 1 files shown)


anacrolix:
Why not `i, t := range`? Can you link to benchmarks showing preallocation is faster than appending to `make([]blah, 0, knownLen)`?

YenForYang:
I think this stackoverflow post discusses this: https://stackoverflow.com/questions/38654729/golang-slice-append-vs-assign-performance

anacrolix:
The SO benchmarks are all from 2016, and don't seem very conclusive (except for some suggestions that append has caught up to assignment in performance).

The torrents map used to be a slice, and I did some shenanigans to make deletions faster, but for some reason that I don't remember it stopped being a performance bottleneck and I moved to using a map.

anacrolix:
I pushed some tests to https://github.com/anacrolix/wat/blob/master/copy_test.go. It's pretty even, enough that I wouldn't care unless it came up in a performance profile (but preallocating is definitely necessary).

PR #565, opened 2021-09-03
acceptLimitClearer: Avoiding reallocating timer in loop

YenForYang:


YenForYang:
Whoops forgot to `Stop()` the timer. I think you can squash the commits

anacrolix:
I think this neglects to Reset the timer or use a ticker.

YenForYang:
Yup you're right; I forgot to account for the `lock()` delay.

anacrolix:
Is it really worth the extra 3 lines for something that allocates once every 15 mins? We'd be better off having the entire goroutine run from a timer, so it's not sitting in the scheduler: except that it's a good way to know if the routine has leaked 🤔 .

YenForYang:
Yep definitely not worth it lol. I missed the simpler, better solution -- using a Ticker. This should be better than using `timer.After`.

anacrolix:
I think I prefer to keep this the way it was. Thanks for experimenting with it.

PR #566, opened 2021-09-04
requester: tweak timer usage

YenForYang:
Conservative timer allocation; use a single timer with Reset instead of reallocating two timers every loop


anacrolix:
I don't think this change is worthwhile.

PR #567, opened 2021-09-04
Create default constructor for Client

YenForYang:
Allow for certain values to always be properly initialized on construction -- namely the maps for now. I'm currently working on a change that requires a baseline constructor; this change would make the use of `chansync.BroadcastCond` and `chansync.SetOnce` obsolete -- i.e. one can have channel members without worrying about proper initialization/destruction of a `chan struct{}`.

As for why `makeClient` returns a value instead of a pointer: returning a value gives us more options -- you can always take a pointer from a value later on cheaply, and have things moved to the heap if they weren't already. The same can't be said about getting a value back from a pointer. GC also could potentially have less work to do. Plus I personally find ownership to be an important concept (semi-borrowed from rust) -- use of values make ownership clear.

~~NOTE: various changes to the tests (in `_test.go` files) still need changing to use this new constructor (instead of their current usage of `Client{...}` or `&Client{...}`)~~

### DIFF
--- a/client.go
+++ b/client.go
@@ -182,32 +182,35 @@ func (cl *Client) announceKey() int32 {
 	return int32(binary.BigEndian.Uint32(cl.peerID[16:20]))
 }
 
+// Initializes a bare minimum Client. *Client and *ClientConfig must not be nil.
+func (cl *Client) init(cfg *ClientConfig) {
+	cl.config = cfg
+	cl.dopplegangerAddrs = make(map[string]struct{})
+	cl.torrents = make(map[metainfo.Hash]*Torrent)
+	cl.dialRateLimiter = rate.NewLimiter(10, 10)
+	cl.activeAnnounceLimiter.SlotsPerKey = 2
+	
+	cl.event.L = cl.locker()
+	cl.ipBlockList = cfg.IPBlocklist
+}
+
 func NewClient(cfg *ClientConfig) (cl *Client, err error) {
 	if cfg == nil {
 		cfg = NewDefaultClientConfig()
 		cfg.ListenPort = 0
 	}
-	defer func() {
-		if err != nil {
-			cl = nil
-		}
-	}()
-	cl = &Client{
-		config:            cfg,
-		dopplegangerAddrs: make(map[string]struct{}),
-		torrents:          make(map[metainfo.Hash]*Torrent),
-		dialRateLimiter:   rate.NewLimiter(10, 10),
-	}
-	cl.activeAnnounceLimiter.SlotsPerKey = 2
+	var client Client
+	client.init(cfg)
+	cl = &client
 	go cl.acceptLimitClearer()
 	cl.initLogger()
 	defer func() {
-		if err == nil {
-			return
+		if err != nil {
+			cl.Close()
+			cl = nil
 		}
-		cl.Close()
 	}()
-	cl.event.L = cl.locker()
+	
 	storageImpl := cfg.DefaultStorage
 	if storageImpl == nil {
 		// We'd use mmap by default but HFS+ doesn't support sparse files.
@@ -220,9 +223,6 @@ func NewClient(cfg *ClientConfig) (cl *Client, err error) {
 		storageImpl = storageImplCloser
 	}
 	cl.defaultStorage = storage.NewClient(storageImpl)
-	if cfg.IPBlocklist != nil {
-		cl.ipBlockList = cfg.IPBlocklist
-	}
 
 	if cfg.PeerID != "" {
 		missinggo.CopyExact(&cl.peerID, cfg.PeerID)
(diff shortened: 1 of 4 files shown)


YenForYang:
Should be good to go now; I updated the tests as well.

YenForYang:
Alright, good to go. I'm hoping to get this commit pushed as it'll make future `Client` changes much easier to manage.

client.go
```diff
@@ -182,32 +182,35 @@ func (cl *Client) announceKey() int32 {
...
 	defer func() {
-		if err == nil {
-			return
+		if err != nil {
+			cl.Close()
+			cl = nil
 		}
-		cl.Close()
 	}()
 	cl.event.L = cl.locker()
```
anacrolix, line 212: Shouldn't this be in makeClient?

anacrolix (commented):
I like this, looks like a few other things can be moved in to makeClient.

client.go
YenForYang: I was getting a build issue when I moved it in makeClient for some reason. There was a panic about a lock / RWMutex ... ah I don't remember exactly but it was frustrating. Anyway it went away right when I moved `cl.event.L = cl.locker()` back to its original position. Probably a deeper issue at hand. I can move it back to see if the issue still exists. Wanna try that?

YenForYang: I'll just do it, and if it comes up -- have a look lol.

YenForYang:
You can check the Go / build failure message -- this is what happens when I move the `locker()` into `makeClient`. Any good workaround? If not we'll just have to dump it to `NewClient` :(

YenForYang:
Found the relevant part of the error:

```
fatal error: sync: Unlock of unlocked RWMutex

goroutine 1256 [running]:
runtime.throw({0x10ab30e, 0x0})
	/opt/hostedtoolcache/go/1.17.0/x64/src/runtime/panic.go:1198 +0x71 fp=0xc00043f4e8 sp=0xc00043f4b8 pc=0x451691
sync.throw({0x10ab30e, 0x488f25})
	/opt/hostedtoolcache/go/1.17.0/x64/src/runtime/panic.go:1184 +0x1e fp=0xc00043f508 sp=0xc00043f4e8 pc=0x48145e
sync.(*RWMutex).Unlock(0xc000141a98)
	/opt/hostedtoolcache/go/1.17.0/x64/src/sync/rwmutex.go:142 +0x7e fp=0xc00043f548 sp=0xc00043f508 pc=0x4972de
github.com/anacrolix/sync.(*RWMutex).Unlock(0xc000141968)
	/home/runner/go/pkg/mod/github.com/anacrolix/sync@v0.4.0/rwmutex.go:26 +0x5b fp=0xc00043f560 sp=0xc00043f548 pc=0x9dd87b
github.com/anacrolix/torrent.(*lockWithDeferreds).Unlock(0xc000141968)
	/home/runner/work/torrent/torrent/deferrwl.go:22 +0xcf fp=0xc00043f598 sp=0xc00043f560 pc=0xcff02f
sync.(*Cond).Wait(0xc000140bc8)
	/opt/hostedtoolcache/go/1.17.0/x64/src/sync/cond.go:55 +0x92 fp=0xc00043f5e0 sp=0xc00043f598 pc=0x493532
github.com/anacrolix/torrent.(*Piece).VerifyData(0xc0004f8600)
	/home/runner/work/torrent/torrent/piece.go:170 +0x225 fp=0xc00043f690 sp=0xc00043f5e0 pc=0xd19be5
github.com/anacrolix/torrent.(*Torrent).VerifyData(0xc000372000)
	/home/runner/work/torrent/torrent/torrent.go:2086 +0x3d fp=0xc00043f6c0 sp=0xc00043f690 pc=0xd3bcfd
github.com/anacrolix/torrent/test.testClientTransfer(0xc00024b6c0, {0x0, 0x2, 0x1, 0xc00060c110, 0x10ca9a0, 0x0, 0x0, {0x0, 0x0}, ...})
	/home/runner/work/torrent/torrent/test/transfer_test.go:96 +0x53a fp=0xc00043fd88 sp=0xc00043f6c0 pc=0xe1c77a
github.com/anacrolix/torrent/test.TestClientTransferVarious.func2.1.1.2(0x0)
	/home/runner/work/torrent/torrent/test/transfer_test.go:343 +0x17d fp=0xc00043fe98 sp=0xc00043fd88 pc=0xe1f49d
```

Looks like the double unlock happens in `(*Piece).VerifyData`, although I'm not sure why yet.

YenForYang:
Oh wow -- just realized Go copies the lock. Now I get why people will return a pointer to struct rather than just a plain struct. Unbelievable, I had no idea Go couldn't even do return value optimization

YenForYang:
Finally got it working. Super frustrating that Go can't do copy elision to avoid `sync.Cond` getting copied. Had to pass explicitly, which makes things more verbose by a line or two.

anacrolix (commented):
My comment didn't go through, would have saved you a lot of time.

anacrolix:
I think I'd rather just have Client.init method now.

YenForYang:
Ah yeah I'll just rename `makeClient` to `initClient` and make it a method if that's okay with you.

anacrolix:
Yes, just `(*Client) init`.

YenForYang:
Alright good to go at last

PR #568, opened 2021-09-04
ListenAddrs: Preallocate slice and make inlineable

YenForYang:


anacrolix:
Not this one, I still prefer append.

PR #571, opened 2021-09-04
Improvements to AnnounceList methods

YenForYang:
- Make `Clone()` preallocate and inlineable; use make+copy optimization since Go 1.15 (https://github.com/golang/go/issues/26252)
- Make `OverridesAnnounce()` inlineable; hoist `announce` check out of loop.
- Preallocate return slice in `DistinctValues()`.

### DIFF
--- a/metainfo/announcelist.go
+++ b/metainfo/announcelist.go
@@ -2,34 +2,46 @@ package metainfo
 
 type AnnounceList [][]string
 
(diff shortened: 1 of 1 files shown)


anacrolix:
I think this one is overkill. It also subtly changes the return of OverridesAnnounce in ways I haven't looked in to.

PR #572, opened 2021-09-04
Drop xerrors and missinggo/slices dependency for Client

YenForYang:
Made `BadPeerIPs` inlineable since I changed `badPeerIPsLocked` anyway. Also eliminated an extra torrentsAsSlice copy since Go wasn't smart enough to figure it out.

### DIFF
--- a/client.go
+++ b/client.go
@@ -142,7 +142,7 @@ func (cl *Client) WriteStatus(_w io.Writer) {
 	})
 	spew.Fdump(w, &cl.stats)
 	torrentsSlice := cl.torrentsAsSlice()
-	fmt.Fprintf(w, "# Torrents: %d\n", len(torrentsAsSlice))
+	fmt.Fprintf(w, "# Torrents: %d\n", len(torrentsSlice))
 	fmt.Fprintln(w)
 	sort.Slice(torrentsSlice, func(l, r int) bool {
 		return torrentsSlice[l].infoHash.AsString() < torrentsSlice[r].infoHash.AsString()
--- a/client.go
+++ b/client.go
@@ -147,8 +147,7 @@ func (cl *Client) WriteStatus(_w io.Writer) {
 	sort.Slice(torrentsSlice, func(l, r int) bool {
 		return torrentsSlice[l].infoHash.AsString() < torrentsSlice[r].infoHash.AsString()
 	})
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -29,7 +29,6 @@ import (
 	"github.com/google/btree"
 	"github.com/pion/datachannel"
 	"golang.org/x/time/rate"
-	"golang.org/x/xerrors"
```
anacrolix, line 32: xerrors is still used elsewhere in teh file

client.go
```diff
@@ -134,11 +141,14 @@ func (cl *Client) WriteStatus(_w io.Writer) {
...
+	torrentsSlice := cl.torrentsAsSlice()
+	fmt.Fprintf(w, "# Torrents: %d\n", len(torrentsSlice))
 	fmt.Fprintln(w)
-	for _, t := range slices.Sort(cl.torrentsAsSlice(), func(l, r *Torrent) bool {
-		return l.InfoHash().AsString() < r.InfoHash().AsString()
-	}).([]*Torrent) {
+	sort.Slice(torrentsSlice, func(l, r int) bool {
+		return torrentsSlice[l].infoHash.AsString() < torrentsSlice[r].infoHash.AsString()
+	})
+	for i := 0; i < len(torrentsSlice); i += 1 {
```
anacrolix, line 150: The rest is good, but just use range syntax here

client.go
```diff
@@ -841,15 +851,13 @@ func (cl *Client) receiveHandshakes(c *PeerConn) (t *Torrent, err error) {
 		err = errors.New("connection does not have required header obfuscation")
 		return
 	}
-	ih, err := cl.connBtHandshake(c, nil)
-	if err != nil {
-		err = xerrors.Errorf("during bt handshake: %w", err)
+	if ih, err := cl.connBtHandshake(c, nil); err == nil {
```
anacrolix, line 854: I think leave this block the way it was.

anacrolix (changes requested):

client.go
YenForYang: I tend to try to reduce the scope of variables whenever possible -- I think I was trying to do that for `ih` here.

I guess I could invert the if the statement to achieve the same effect of returning the error first.

PR #573, opened 2021-09-04
listenAll: Preallocate []networkAndHost and make inlineable

YenForYang:
Might as well -- never mind that listenAll isn't called a whole bunch.


anacrolix:
I think this one is overkill.

PR #574, opened 2021-09-07
AddClientPeer: Preallocate []PeerInfo and make inlineable

YenForYang:


anacrolix:
I know the functional style is overkill, but I still prefer append unless we know that assignment will perform better in the specific case.

PR #575, opened 2021-09-08
SetMaxEstablishedConns: Predetermine dropped connection count

YenForYang:
Also build []*PeerConn without reflection.


anacrolix:
This seems like overkill, but I'm curious about the suggestion for a better algorithm.

PR #576, opened 2021-09-08
readOnceAt: Move loop-invariant code out of loop body

YenForYang:
Separate function into an outlined "slow" variant and inlineable "fast" variant. The assumption is that errors reading a torrent are relatively rare, so looping and locking are generally not required.


anacrolix:
I'm not sure what this achieves? Do you hope to inline the fast path?

PR #577, opened 2021-09-08
Inlineable reader.Seek and no-lock return for bad whence

YenForYang:
Couldn't find an elegant way to early exit on bad whence. Using a closure could reduce code duplication slightly, but it's overkill for only 3 cases.

Note that returning 0 on an invalid whence is the behavior of `strings.Reader` and `bytes.Reader`.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -250,24 +250,27 @@ func (r *reader) posChanged() {
 }
 
 func (r *reader) Seek(off int64, whence int) (ret int64, err error) {
-	r.opMu.Lock()
-	defer r.opMu.Unlock()
-
-	r.mu.Lock()
-	defer r.mu.Unlock()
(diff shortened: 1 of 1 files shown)


YenForYang:
Dont know if the extra commits can be squashed since I merged -- if not I can close this request and open a new pull

anacrolix:
Looks good. Have you compared the assembly side by side? I wonder if having locks in different locations actually blows out the function size.

YenForYang:
Hmm I have not. Have there been issues in the past like that in Go? I did a crude check on [godbolt](https://godbolt.org/z/s6YvoMjG5) (I took out `posChanged()` to remove the *Torrent dependency). Didn't look like anything exploded lol

PR #578, opened 2021-09-08
Update methods to support pointer receiver

YenForYang:
`(*metainfo.Hash).HexString()` currently uses a value receiver, which allows for chaining method calls (e.g. `.Hash().HexString()`). This is nice, but I've been doing some testing with pointer receiver methods -- such methods disallow can't be called on immediately returned values.

To maintain compatibility with both pointer and value receiver methods, a local variable needs to be created. This shouldn't be a big deal, but it helps a lot with code refactoring.

### DIFF
--- a/cmd/torrent-infohash/main.go
+++ b/cmd/torrent-infohash/main.go
@@ -20,6 +20,7 @@ func main() {
 		if err != nil {
 			log.Fatal(err)
 		}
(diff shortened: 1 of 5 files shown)


anacrolix:
Can you address the unit CI failure?

YenForYang:
I'll go ahead and close this request and just work from https://github.com/anacrolix/torrent/pull/579. This will make things easier for me so hope that's okay

PR #579, opened 2021-09-09
Rework metainfo.Hash

YenForYang:
Various changes. Changed methods to use pointer receiver consistently. Most changes are around reducing unnecessary copies and allocations. Also reordered functions into interface implementations, public methods, and then private ones. Hopefully this isn't a problem. Use of `unsafe` is explained in comments. Changed HashBytes to call `sha1.Sum` directly, and more.

~~Requires https://github.com/anacrolix/torrent/pull/578 in order to change methods to be compatible with pointer receivers.~~

### DIFF
--- a/cmd/torrent-infohash/main.go
+++ b/cmd/torrent-infohash/main.go
@@ -20,6 +20,7 @@ func main() {
 		if err != nil {
 			log.Fatal(err)
 		}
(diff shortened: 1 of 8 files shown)


anacrolix:
I'll take another look after the other PR is done.

anacrolix:
This is still an interesting PR, but there are conflicts now. Do you still want to proceed with this one?

PR #580, opened 2021-09-09
(*Torrent).name(): Earlier RUnlock and make inlineable

YenForYang:


anacrolix:
I don't think this is worthwhile.

PR #581, opened 2021-09-09
Simplify (*Torrent).gotMetainfo

YenForYang:
Data races shouldn't be an issue since `missinggo.Event` didn't originally use a lock anyway. As far as I can tell `missinggo.Event` simply provides insurance that the channel isn't closed more than once. If we are careful about keeping the channel initialized on the construction of the torrent, as well as making sure to re-`make` the `gotMetainfo` channel when setting `.info = nil`, then the double close issue shouldn't be encountered.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1111,6 +1111,7 @@ func (cl *Client) newTorrent(ih metainfo.Hash, specStorage storage.ClientImpl) (
 			L: cl.locker(),
 		},
 		webSeeds: make(map[string]*Peer),
+		gotMetainfoC: make(chan struct{}),
 	}
 	t.networkingEnabled.Set()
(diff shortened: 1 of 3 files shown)


anacrolix:
My replacement for missinggo.Event is chansync.SetOnce. https://pkg.go.dev/github.com/anacrolix/chansync

YenForYang:
Hmm but that doesn't seem like that's necessary here. SetOnce appears to have even more precautions than Event. Why not just use a simple `chan struct{}`? I'll probably make a constructor for torrent like I did for `Client` (e.g. `makeTorrent`) to ensure that things are properly initialized in a clearer way.

PR #586, opened 2021-09-10
(*Client).requester() rework

YenForYang:
Cut down from 2 timers reallocated each loop to a single timer with proper Reset's and Stop's. Also double check `cl.closed` before making requests.

### DIFF
--- a/requesting.go
+++ b/requesting.go
@@ -1,9 +1,10 @@
 package torrent
 
 import (
+	"math"
 	"time"
 	"unsafe"
-
+	
 	"github.com/anacrolix/missinggo/v2/bitmap"
 
 	"github.com/anacrolix/chansync"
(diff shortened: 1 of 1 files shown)


anacrolix:
I really think it's better to use the simple version of these unless we know they're a string performance impact. The timer allocation issue comes up so much it's probably worth making an abstraction that avoids the allocation but retains the easy interface.

YenForYang:
It does add some cognitive overhead without a doubt, so I understand the concern. But in this case it does actually make a difference (compared to `acceptLimitClearer`, which only happened once every 15 minutes).

I made an effort to make things clear, but if my code is hard to follow I can make changes. I saw your post on https://github.com/golang/go/issues/14383 while researching correct use of timers. `timer.Stop` is definitely a pain to use correctly. Honestly, `time.After` is probably the only intuitive way to use a Timer lol -- considering how they are mostly 1 time use things. And yet it Go doesn't bother optimizing idioms involving re-using them in a loop; you have to assume that the GC will correctly stop and free the Timer.

anacrolix:
Yeah if you can come up with an abstraction around time.After that avoids reallocating and guarantees no races, I'll use that, since it would be just one allocation at the start of the loop (us asserting that we know it's okay to reuse the instance), and changing time.After to `preallocatedTimer.After` or whatever the new API is.

Issue #534, opened 2021-07-26
Url-encoded `info_hash` uses `+` instead of `%20`


YenForYang:
I discovered a problem when attempting to download a torrent from a private tracker. I couldn't figure out why things weren't working: the torrent worked fine under Transmission, qBittorrent, etc., but hung indefinitely at 0 using Go. I turned on debug mode and got some interesting messages:

```
2021-07-26T08:14:48+0200 DEBUG client.go:901: received handshake for unloaded torrent
...
2021-07-26T08:14:48+0200 DEBUG client.go:892: error receiving handshakes on connection 0xc0000b8700: error while establishing secret: error reading Y: unexpected EOF
2021-07-26T08:15:31+0200 DEBUG tracker_scraper.go:145: "...": announcing to "http://.../announce": udp.AnnounceRequest{InfoHash:[20]uint8{0xb7, 0x76, 0xa, 0xa1, 0x78, 0x93, 0x20, 0x30, 0xc8, 0x47, 0xdc, 0xdf, 0x8e, 0xae, 0xbf, 0x56, 0xa, 0x1b, 0xd1, 0x6c}, PeerId:[20]uint8{...}, Downloaded:0, ...}
2021-07-26T08:15:31+0200 DEBUG tracker_scraper.go:158: "...": announce to "http://.../announce" returned http.AnnounceResponse{Interval:0, Leechers:0, Seeders:0, Peers:[]http.Peer(nil)}: tracker gave failure reason: "Unregistered torrent"
...
2021-07-26T08:16:10+0200 DEBUG client.go:901: received handshake for unloaded torrent
...
2021-07-26T08:16:19+0200 DEBUG client.go:901: received handshake for unloaded torrent
...
2021-07-26T08:16:20+0200 DEBUG client.go:892: error receiving handshakes on connection 0xc0000b8700: error while establishing secret: error reading Y: unexpected EOF
2021-07-26T08:16:31+0200 DEBUG tracker_scraper.go:145: "...": announcing to "http://.../announce": udp.AnnounceRequest{InfoHash:[20]uint8{0xb7, 0x76, 0xa, 0xa1, 0x78, 0x93, 0x20, 0x30, 0xc8, 0x47, 0xdc, 0xdf, 0x8e, 0xae, 0xbf, 0x56, 0xa, 0x1b, 0xd1, 0x6c}, PeerId:[20]uint8{...}, Downloaded:0, ...}
[...]

PR #591, opened 2021-09-11
Fix http announce of infohash containing ' ' bytes

anacrolix:
Fixes https://github.com/anacrolix/torrent/issues/534.

### DIFF
--- a/tracker/http/http.go
+++ b/tracker/http/http.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


YenForYang:
Yeah this is what I thought originally would work, but then it came to me that maybe (now or in the future) there would be `+` signs that shouldn't be changed to `%20`. In other words, we probably shouldn't be trying to check the entire result to avoid future issues like this.

anacrolix:
I think we can do it this way: hard to imagine that only info_hash is being decoded in the way we described. There could be trackers out there that expect `+` instead but it seems unlikely ($50 it's something Amazon made if there is).

PR #592, opened 2021-09-11
Optimize getIp() and make inlineable

YenForYang:
A few changes that make things clearer and reduce chances of hitting another Go gotcha. Go also is pretty bad at moving loop-invariant code out of loops, so this should help thing in that front as well.

### DIFF
--- a/tracker_scraper.go
+++ b/tracker_scraper.go
@@ -65,10 +65,12 @@ type trackerAnnounceResult struct {
 	Completed time.Time
 }
 
-func (me *trackerScraper) getIp() (ip net.IP, err error) {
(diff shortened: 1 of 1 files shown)


anacrolix:
Please use go fmt.

YenForYang:
Usually my formatting lines up exactly with go fmt, but when it comes to `switch` I get carried away with making the lines line up lol. I fixed it.

anacrolix:
I think Go's early return convention still has legs here. This function accumulates indentation in the new form.

YenForYang:
I could move the variable declaration out of the switch and remove the default case to reduce indentation, but that would increase variable scope. I personally would rather reduce scope than worry about indentation. But I'm willing to change it if you have any ideas.

PR #593, opened 2021-09-11
NewClient: Better panic messages for PeerID creation

YenForYang:
In the future, it may be a good idea to change `ClientConfig.PeerID` to have a `*PeerID` type or `*[20]byte` type. It would make error checking easier and make it clear that PeerID is either to be empty or 20 bytes long.

I would change it, but it looks like `.PeerID` is a public field -- changing would probably break existing programs.

### DIFF
--- a/client.go
+++ b/client.go
@@ -232,14 +232,16 @@ func NewClient(cfg *ClientConfig) (cl *Client, err error) {
 		cl.ipBlockList = cfg.IPBlocklist
 	}
 
-	if cfg.PeerID != "" {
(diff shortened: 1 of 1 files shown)


anacrolix:
This subtly changes the behaviour, it now ignores cfg.PeerId if it's the wrong length. Some of my APIs do intentionally not use pointers for performance reasons, like in anacrolix/dht/v2/krpc, but that is not the case here.

YenForYang:
Huh if it does change the behavior then I made a mistake; I'll look into it. I thought `missinggo.CopyExact` just panics if destination and source lengths don't match, right?

PR #594, opened 2021-09-12
Inlineable (*reader).SetReadAhead and (*reader).Close

YenForYang:
Small fixes that prevent inlining of public functions

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -69,10 +69,10 @@ func (r *reader) SetNonResponsive() {
 
 func (r *reader) SetReadahead(readahead int64) {
 	r.mu.Lock()
-	defer r.mu.Unlock()
 	r.readahead = readahead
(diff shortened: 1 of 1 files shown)


anacrolix:
I couldn't find evidence of this using `-gcflags -m`. That is, anacrolix/confluence, a user of the public anacrolix/torrent interface, doesn't inline SetReadahead whether it's defer or not. If we can set a precedent that defers do block inlining, I'm happy to accept more PRs for them without feedback.

YenForYang:
There's always the chance that the function cost exceeds the budget of 80 (a number hardcoded by Go currently). But future changes could potentially reduce the cost, allowing the functions to be inlined. On the other hand, `defer` blocks the function from being inlined regardless of how big or small it is.

PR #598, opened 2021-09-12
Pool readers and create default constructor

YenForYang:
I was hesitant to make this request before, but it now seems like a pretty obvious choice as the `reader` struct begins to get more fields. Torrents with thousands of files would easily benefit from this. On the hand, if a user only creates a single reader for a torrent -- there would likely be no benefit.

But the added bonus is that it goes along nicely with constructors! The constructor here is obviously `newReader`

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -61,7 +61,6 @@ var readerPool = sync.Pool {
 func newReader(t *Torrent,
 	offset, length, pos, readahead int64,
 	responsive bool,
-	readAheadFunc func() int64,
 ) (r *reader) {
 	r = readerPool.Get().(*reader)
 	r.t = t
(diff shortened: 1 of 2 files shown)


YenForYang:
I should mention that this pull request incorporates https://github.com/anacrolix/torrent/pull/594, so that one can be closed when this one is accepted.

YenForYang:
Looks like https://github.com/anacrolix/torrent/pull/594 was merged, so I made some updates.

anacrolix:
Is there some evidence this is a performance bottleneck to begin with?

YenForYang:
The performance improvement can't be easily proven here without actually testing some actual torrents with lots of files. But this pull request is less about pure speed and more about efficiency and reducing allocations / GC workload.

The case for `sync.Pool` here is like the case for using a `sync.Pool` in `fmt` -- currently the `print`/`printf`-like functions in `fmt` pool all their formatted strings in `ppFree`. `fmt` even pools their scanners (e.g. `Fscan`, `Fscanf`, etc.) in a variable named `ssFree`. I doubt most individuals will be formatting/scanners enough strings to see a measurable advantage over not using `sync.Pool`, yet the `sync.Pool` usage still exists just because there is a good chance the types will be used a lot.

Ultimately I think pooling `reader` is a really good idea, especially as we aim to add/improve functionality of `reader`. I can't offer hard evidence here unfortunately, but I'm hoping this is enough to clear most doubts.

anacrolix:
Consider that we can't even guarantee users of reader will stop using it. And if we created a wrapper type that detached itself from the existing reader, it would require an allocation to be able to detach it. I don't know that this is worth it, as clean as the code change is. Just calling addReader would cost so much more than any allocation of *reader.

anacrolix:
Nice PR but as I mentioned above I'm not sure there's any performance issues here.

PR #601, opened 2021-09-12
Inlineable methods to access File piecePriority

YenForYang:
Seems worthwhile when setting/getting priorities for many files

### DIFF
--- a/file.go
+++ b/file.go
@@ -146,19 +146,19 @@ func (f *File) NewReader() Reader {
 // Sets the minimum priority for pieces in the File.
 func (f *File) SetPriority(prio piecePriority) {
 	f.t.cl.lock()
-	defer f.t.cl.unlock()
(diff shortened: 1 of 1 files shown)


anacrolix:
Do you have some evidence that defer blocks inlining?

YenForYang:
I simple test on godbolt with `-m=2` would show that `ODEFER` blocks inlining. However, you can also take a look at the source; these lines in particular: https://github.com/golang/go/blob/ad97d204f02c1f9ad0433e9178d6ce0f3fdb1f9f/src/cmd/compile/internal/inline/inl.go#L344-L351.

Note that `range`, `select`, `go`, `defer`, tail calls, and `type` block inlining in a function.

anacrolix:
Thanks for the thorough walkthrough. Go sux.

PR #602, opened 2021-09-13
Cheaper byte to string conversion

YenForYang:
Revamped https://github.com/anacrolix/torrent/pull/587


anacrolix:
This is the same thing you had before. But I read https://stackoverflow.com/questions/59209493/how-to-use-unsafe-get-a-byte-slice-from-a-string-without-memory-copy/59210739#59210739. Let's give it a go.

PR #603, opened 2021-09-13
lockWithDeferreds: inlineable Unlock()

YenForYang:


anacrolix:
Seems unnecessary. Is there some cost to accessing the actions through ranging over me.unlockActions?

YenForYang:
No idea, but it seems like `(*lockWithDeferreds).Unlock()/Lock()` functions are used pretty often (e.g. `(*Client).lock()/unlock()`). At the very least this would allow `cl._mu.Unlock()` to be inlined inside the one-liner function `(*Client).unlock()`.

anacrolix:
The inlining seems worth it, thanks.

PR #604, opened 2021-09-14
Refactor `types` (Request, ChunkSpec, piecePriority)

YenForYang:
Time to start getting rid of those type aliases; use them for gradual code refactoring.


YenForYang:
Not sure if you intend to keep them exported. Might as well just use the `types` package and not bloat the code -- but I don't know, up to you.

anacrolix:
This has drifted too far from master. I don't intend to make this change as it is.

PR #606, opened 2021-09-14
Create helper-type `readers` to hold `*reader`

YenForYang:
Currently `*Torrent` uses a `map[*reader]struct{}` to store `reader`, which is unnecessary. A slice is all that's needed. We just need to add a field (e.g. `idx`) to `*reader` in order to mimic the `delete` functionality maps have.

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -48,6 +48,9 @@ type reader struct {
 	// required.
 	pieces pieceRange
 
+	// Internal identifier used to index a []*reader
+	idx uint32
+	
 	// Reads have been initiated since the last seek. This is used to prevent readaheads occurring
 	// after a seek or with a new reader at the starting position.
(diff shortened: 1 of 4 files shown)


anacrolix:
This would actually be way more interesting on Client.torrents, where would probably see some performance gains.

YenForYang:
Yep it definitely would, but currently `Client.torrents` uses a `map[InfoHash]*Torrent`, so the issue would be moving away from using `InfoHash` to a slice index. Slightly more involved but probably worth it to remove the infohash requirement to access a torrent.

YenForYang:
Any issues with this PR?

anacrolix:
I would like to see that the existing code is a performance bottleneck first before complicating this.

PR #608, opened 2021-09-14
Simplify `(*Torrent).wantPeersEvent`

YenForYang:
Make `wantPeersEvent` a simple `chan struct{}`.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1114,6 +1114,7 @@ func (cl *Client) newTorrent(ih metainfo.Hash, specStorage storage.ClientImpl) (
 		},
 		webSeeds: make(map[string]*Peer),
 		gotMetainfoC: make(chan struct{}),
+		wantPeersEvent: new(chan struct{}),
 	}
 	t.networkingEnabled.Set()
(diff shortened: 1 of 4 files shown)


YenForYang:
I'm thinking about maybe making a small wrapper type to to encapsulate the `chan struct{}`, which will probably make this more readable. Something like `type ChanEmptyStruct chan struct{}`.  Probably will make an atomic version as well.

YenForYang:
Alright should be good now. Didn't think too hard about naming the methods for the wrapper -- I've seen stuff like `Broadcast()`, `Signal()`, `Set()`, etc. but since all that's up for debate I just used `Close`, `Reset()`, etc. to state what happens to the channel rather than try to create a more abstract perspective. Definitely was a pain to write this boilerplate.

Combined with the new client initializer, we can change the types of a couple other fields to eliminate a few mutexes and/or `sync.Once`'s

anacrolix:
I think I have anacrolix/chansync for this kind of thing now. It is used elsewhere in the codebase, and can be used here when there's reason to upgrade.

PR #609, opened 2021-09-14
Clarify maximum value of "metadata_size"

YenForYang:
Not saying that https://github.com/arvidn/libtorrent has all the answers, but I think it's a good reference: https://github.com/arvidn/libtorrent/blob/95afba30e9acfd8411076a0b05f7ca69bf33b9bb/src/ut_metadata.cpp#L145-L151.

They use `4 * 1024 * 1024`, but we can double that to be safe. Note how they also use a multiple of 16*1024. I found another torrent client that used `2 * 1024 * 1024`, so we are on the high end regardless.

---

Had a nice chuckle at the comment next to 10MB.

Fixed a redundant check and future possible clash with `bytes` package name.

### DIFF
--- a/torrent.go
+++ b/torrent.go
@@ -481,19 +481,19 @@ func (t *Torrent) haveAllMetadataPieces() bool {
 }
 
 // TODO: Propagate errors to disconnect peer.
-func (t *Torrent) setMetadataSize(bytes int) (err error) {
+func (t *Torrent) setMetadataSize(size int) (err error) {
 	if t.haveInfo() {
 		// We already know the correct metadata size.
 		return
 	}
-	if bytes <= 0 || bytes > 10000000 { // 10MB, pulled from my ass.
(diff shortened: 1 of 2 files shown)


torrent.go
```diff
@@ -481,19 +481,19 @@ func (t *Torrent) haveAllMetadataPieces() bool {
...
 
 // TODO: Propagate errors to disconnect peer.
-func (t *Torrent) setMetadataSize(bytes int) (err error) {
+func (t *Torrent) setMetadataSize(size int) (err error) {
 	if t.haveInfo() {
 		// We already know the correct metadata size.
 		return
 	}
-	if bytes <= 0 || bytes > 10000000 { // 10MB, pulled from my ass.
+	if uint32(size) > maxMetadataSize {
```
anacrolix, line 489: It's the end of an era. There are plenty more comments like these 😆

PR #611, opened 2021-09-14
Use ContextDialer from official Go subrepo

YenForYang:
Seems like "golang.org/x/net" is a project dependency anyway.


anacrolix:
It's our interface, I'd rather not confuse it with the proxy one. If proxy.Dialer also implements Network though, I'm okay with it, since it would communicate that a proxy.Dialer is an appropriate instance for the interface.

PR #614, opened 2021-09-14
Simplify `(*Torrent).maxHalfOpen()`

YenForYang:
Clean up some unnecessary functions that weren't inlined. `min` and `minInt` remain unused.


anacrolix:
I prefer the older version. You can make generic implementations of min and max now that would remove overhead.

PR #616, opened 2021-09-14
Correct `connIsIpv6`

YenForYang:
Not sure if the current implementation works, or if it does what it's intended to do. Looks like it checks if an address is IPv6 -- if this is the case, then there is a less roundabout way of doing so.

### DIFF
--- a/client.go
+++ b/client.go
@@ -382,7 +382,7 @@ func (cl *Client) NewAnacrolixDhtServer(conn net.PacketConn) (s *dht.Server, err
 		Conn:           conn,
 		OnAnnouncePeer: cl.onDHTAnnouncePeer,
 		PublicIP: func() net.IP {
-			if connIsIpv6(conn) && cl.config.PublicIp6 != nil {
+			if connIsIpv6(conn.LocalAddr()) && cl.config.PublicIp6 != nil {
 				return cl.config.PublicIp6
 			}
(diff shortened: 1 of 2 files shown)


anacrolix:
I've had issues with net.Addr.String allocating too much in the past. I'm worried about changing this.

tsynik:
return rip.To4() == nil && rip.To16() != nil works as expected.

YenForYang:
Hmm then I'll change it to not use `String` if the current implementation works as expected. Namely the `To4()` and `To16()` calls can be squashed into something less confusing, but slightly more verbose.

YenForYang:
@anacrolix `addrIpOrNil` was used in the original implementation and also calls `String()` if `net.Addr` is not `UDPAddr` or `TCPAddr`. I think I'll fallback to a string check just like `addrIpOrNil`.

tsynik:
this eliminates use of net.Addr.String but doesn't it better use  addrIsIpv6 for new func?

anacrolix:
It's an interesting idea, and I like your intrusive ipv4 in ipv6 test, but I'm loathe to change this without more tests and benchmarks. Especially regarding the connLessTrusted call, which I think is called in to from somewhere performance critical at the moment.

anacrolix:
If I change this I'll probably use `net/netip`, and propagate the family around with the connection.

PR #617, opened 2021-09-14
Change `Private` field of `metainfo.Info` to `bool`

YenForYang:
Not sure why a pointer is used here since none of the other fields are pointers -- if "private" doesn't exist it's false as expected. This also saves a `nil` check.


anacrolix:
I suspect it has to do with ensuring that marshalling an unmarshalled Info that did previously contain "private" does not change the infohash. It might take an extended check to make this change. Of course Go could just have an Optional/Maybe type and this wouldn't be necessary. Maybe we can wait for that.

anacrolix:
It's now possible to use generics for this, if it bothers anyone.

PR #619, opened 2021-09-14
Revise `nominalMaxRequests()`

YenForYang:
Using a `switch` instead of a clamp function here is ultimately more readable and efficient. I put `default` first because maybe someday Go will make it useful for branch prediction (they allow default to be anywhere in the switch statement -- I can't think of any better reason why(.


anacrolix:
I prefer it the way it was, cheers.

PR #631, opened 2021-09-15
Optimize `initFiles()` and early cache `length`

YenForYang:
`initFiles()` is now split to perform differently depending on whether the torrent contains multiple files. I removed the `range` and it shouldn't affect readability much in this case; although, with the large size of the function, I doubt `initFiles()` can be inlined even without `range` inside. 

There was some confusing use of `path`, `.PathUTF8`, `.Path` inside the original subroutine, so I changed `initFiles` to pick between `PathUTF8` and `Path`, store that in `path`, and then use `path` to make things clearer. `cacheLength()` is now obsolete.

### DIFF
--- a/t.go
+++ b/t.go
@@ -210,32 +210,53 @@ func (t *Torrent) cancelPiecesLocked(begin, end pieceIndex) {
 	}
 }
 
+// Initialize files and cache torrent length
 func (t *Torrent) initFiles() {
-	var offset int64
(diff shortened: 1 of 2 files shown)


anacrolix:
I like the `PathUTF8` handling improvements, and the fixes to use `%w`. I don't think the rest is necessary. Please consider resubmitting.

PR #633, opened 2021-09-15
Inlineable `(*Torrent).BytesMissing()`

YenForYang:
Honestly a name like `BytesLeft` would have been more suitable, but it's too late for that I guess.


YenForYang:
Any changes needed here?

anacrolix:
I think missing was chosen instead of left because the user may not be trying to download everything.

PR #642, opened 2021-09-15
Extract `maxRequests` from `(*Torrent).addWebSeed()`

YenForYang:
Just some basic code refactoring; extract the constant to use it in `WebseedHttpClient` as well (I'm assuming `MaxConnsPerHost` should have the same value.


anacrolix:
I think they're the same value by coincidence. I don't think they should be linked.

anacrolix:
It was a good observation, I think the code has moved on (and the assumptions here aren't correct anymore).

PR #643, opened 2021-09-15
Inlineable `(*Torrent).addPeers()`

YenForYang:


anacrolix:
I prefer the old style.

PR #644, opened 2021-09-15
Change `trackerScraper.lastAnnounce` to `atomic.Value`

YenForYang:
Temporary stand-in until we get something better (hopefully when generics come around).

Also fixed `(trackerScraper).URL()` to use a pointer receiver instead; don't know if original copying was intended (as it wasn't documented).

### DIFF
--- a/tracker_scraper.go
+++ b/tracker_scraper.go
@@ -1,12 +1,16 @@
 package torrent
 
 import (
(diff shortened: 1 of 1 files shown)


anacrolix:
This is out of date. I don't think I was convinced that it was worthwhile.

PR #646, opened 2021-09-15
Optimize (*Torrent).peersAsSlice()

YenForYang:
Probably a controversial change; not expecting it to go in. But it does perform much better nonetheless -- it's just not used enough to be noticeable. I'm okay with leaving it out mainly because I don't use `writeStatus()` at the moment lol.


anacrolix:
Yeah this probably isn't worthwhile. If iterPeers became a bottleneck I would definitely notice in my testing.

PR #647, opened 2021-09-15
Move `AnnounceEvent` to tracker/shared

YenForYang:
Changing this will probably help to eliminate the confusing type aliases and cleanup imports a bit. I don't think `AnnounceEvent` is best suited to be in tracker/udp.

I used golang.org/x/tools/cmd/stringer to generate the `String()` method, which is why it looks weird.


anacrolix:
I think I may have fixed this up, or done more work to improve the situation (since adding tracker server implementations).

PR #650, opened 2021-09-17
bencode: optimize byte slice clone

YenForYang:
See https://github.com/go101/go101/wiki/There-is-not-a-perfect-way-to-clone-slices-in-Go.

### DIFF
--- a/bencode/bytes.go
+++ b/bencode/bytes.go
@@ -8,9 +8,10 @@ var (
 	_ Marshaler   = Bytes{}
 )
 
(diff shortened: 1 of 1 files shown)


anacrolix:
This could be worthwhile, I expect Bytes is used heavily in a few spots. Are there any benchmarks for this for real-world cases? metainfo.Info, and peer_protocol.Msg, and possibly some krpc stuff in anacroilx/dht would be very important ones.

anacrolix:
I think we talked about this elsewhere, I don't think the speedup is worthwhile.

PR #653, opened 2021-09-17
bencode: simplify scanner

YenForYang:


anacrolix:
I think that embedding interfaces is tricky as some functions like io.Copy will find optional interfaces that weren't intended to be exposed directly to the user, thereby bypassing things like the offset counting in the scanner. Let me know if that's incorrect.

PR #654, opened 2021-09-17
bencode: optimize error types in api

YenForYang:
Using `fmt` generally incurs some overhead -- i.e. extra memory to store the format string and extra time to parse it. I wouldn't use it unless there's a clear readability advantage.
Furthermore, it looks like `fmt` is already not used in most cases here; let's make it consistent and change the remaining couple cases.

Note: `(*bytes).Reader.Len()` always returns a value greater than or equal to 0, so the number will be non-negative, allowing `FormatUint` to be used.

Other cosmetic changes are just to appease Go's inliner and make things look cheaper (it's dumb and perceives costs differently.) Hope that isn't an issue.

### DIFF
--- a/bencode/api.go
+++ b/bencode/api.go
@@ -2,9 +2,9 @@ package bencode
 
 import (
 	"bytes"
-	"fmt"
 	"io"
 	"reflect"
+	"strconv"
 
 	"github.com/anacrolix/missinggo/expect"
 )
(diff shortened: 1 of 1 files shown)


anacrolix:
I wouldn't want to change this to a less convenient style unless a benchmark showed it was worthwhile.

PR #658, opened 2021-09-17
bencode: optimize `(*Decoder).parseIntInterface()`

YenForYang:


anacrolix:
The code has moved on, this is done a bit differently now. Benchmarks would definitely be welcomed here tho.

PR #660, opened 2021-09-17
bencode: optimize `(*Decoder).parseDictInterface()`

YenForYang:


anacrolix:
It looks tidier but I don't like the extra parseValueInterface before the loop.

anacrolix:
I'll close this now, there's a conflict and the code has moved on a bit.

PR #662, opened 2021-09-18
metainfo: optimize `(*UrlList).UnmarshalBencode()`

YenForYang:


metainfo/urllist.go
```diff
@@ -10,18 +10,13 @@ var (
 	_ bencode.Unmarshaler = (*UrlList)(nil)
 )
 
-func (me *UrlList) UnmarshalBencode(b []byte) error {
-	if len(b) == 0 {
-		return nil
+func (me *UrlList) UnmarshalBencode(b []byte) (err error) {
+	if len(b) < 1 {
```
anacrolix, line 14: I think you've technically changed it to not fail on single byte inputs. This would be much more interesting as a fuzz target or something.

PR #663, opened 2021-09-18
metainfo: optimize `(*FileInfo).Offset()` and make inlineable

YenForYang:


anacrolix:
I prefer the old style here.

PR #664, opened 2021-09-18
iplist: optimize `IPNetLast()`

YenForYang:
https://go.godbolt.org/z/rnMec6Tnx

---
Yup, Go's compiler is really dumb.


anacrolix:
I'm not sure what this PR accomplishes. But you might be interested in `net/netip` coming in Go 1.18.

anacrolix:
The godbolt looks promising, but I can't even recall what this function is supposed to do 😆

anacrolix:
It should be done with a Benchmark attached. I expect that this only is relevant with some of those really big blocklists, they got up to ~15-20MB at one point.

PR #668, opened 2021-09-23
Enumerate valid cases for `(*network).String()`

YenForYang:
Clearer than if statements that effectively have `fallthrough`. Also limits valid cases.

### DIFF
--- a/networks.go
+++ b/networks.go
@@ -17,15 +17,21 @@ type network struct {
 }
 
 func (n network) String() (ret string) {
-	a := func(b bool, s string) {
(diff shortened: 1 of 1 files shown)


YenForYang:
I can add `n.Ipv4` and `n.Ipv6` as individual cases if desired. But I don't think returning `4` and `6` alone would be intentional.

anacrolix:
I'm not sure about this. The old version is definitely a garbage in, garbage out situation. Go clearly lacks enums or something here.

PR #669, opened 2021-09-23
Simplify `peerNetworkEnabled()`

YenForYang:
Primarily a cosmetic change, but also reduces inline cost slightly.


anacrolix:
I prefer the old version.

PR #670, opened 2021-09-23
Optimize `parseNetworkString()`

YenForYang:
Eliminate function call to `strings.Index` and possible unwanted cases.


anacrolix:
I prefer the old version, I don't think this is performance sensitive.

PR #671, opened 2021-09-24
Manually inline initialization of `allPeerNetworks`

YenForYang:
Go can't inline the function. Additionally removes unnecessary allocations.


anacrolix:
I prefer the old version.

PR #672, opened 2021-09-24
Optimize `(*pexMsgFactory).DeltaLen()`

YenForYang:
Also eliminates verbose type casts


anacrolix:
I'm happy to wait for a generics solution to this.

PR #674, opened 2021-09-24
Use explicitly sized type for `pexEventType`

YenForYang:
Also reordered fields in struct `pexEvent` to move padding to end. Note `pp.PexPeerFlags` is also byte sized.

### DIFF
--- a/pex.go
+++ b/pex.go
@@ -9,7 +9,7 @@ import (
 	pp "github.com/anacrolix/torrent/peer_protocol"
 )
 
-type pexEventType int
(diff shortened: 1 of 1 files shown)


yarikk:
Moving the field breaks the code which uses type literals to initialise values without field names.

I'm not sure what's the motivation behind this change: `int` uses native machine word length; `int8` is explicitly smaller and may incur extra delays upon accesses, especially on RISC platforms. Unless there are clear benefits, I'd prefer to keep this as-is.

PR #675, opened 2021-09-27
Add metadata field to metainfo

DamiaPoquet:
This pull request adds support for the "metadata" filed in the torrent's metainfo.

This field is not officially into any BEP, but it is commonly used by some private trackers to store additional information.

### DIFF
--- a/metainfo/metadata.go
+++ b/metainfo/metadata.go
@@ -1,0 +1,8 @@
+package metainfo
+
+type Metadata struct {
+	CoverUrl    string   `bencode:"cover url,omitempty"`
+	Description string   `bencode:"description,omitempty"`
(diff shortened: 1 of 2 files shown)


DamiaPoquet:
I could not find any public information on that "metadata" field. It is used in private forks of Luminance/Gazelle, so there is nothing publicly available.

As it does not follow any BEP, I understand this code not being merged to your repo. I can continue using my fork.

anacrolix:
Alternatively, there might be something somewhere that says user-defined stuff can be packed in to the metainfo too. If that's the case, there should be a way to access those. See the Params field in https://pkg.go.dev/github.com/anacrolix/torrent/metainfo#Magnet for example.

anacrolix:
This was resolved by using a fork.

PR #677, opened 2021-10-10
Add `DisableInitialPieceCheck` option

skirsten:
This greatly reduces the startup time of large torrents, during which a bunch of zero pieces get hashed.

For a huge media torrent that requires a seek to the end for playback, this removes ~3 minutes of waiting for me.﻿


anacrolix:
Out of curiosity, how is it for torrents you've used before, due to the piece completion database?

Issue #691, opened 2021-12-06
Memory use by PEX grows significantly


anacrolix:
In a very long-running client instance where Torrents are not closed, memory use due to PEX grows quite large. `pexState.ev`. 

`inuse_space`:

![image](https://user-images.githubusercontent.com/988750/144789397-bf3a2208-9dcc-4d32-a6d7-be76a01eb3b0.png)

`inuse_objects`:

![image](https://user-images.githubusercontent.com/988750/144789470-ca6d9aa2-a4b4-4cca-8310-23fafe5eab8b.png)

Any chance you can take a look @yarikk? I expect some condition where the append-only log can be trimmed will fix the problem.

yarikk:
Yes, I think I have a few ideas to play with.

yarikk:
I can think of two approaches, each with its pros and cons:
1) periodically shift-and-truncate `t.pex.ev` based on minimal values for `t.conn[*].pex.seq`, tracking the virtual offset after truncations.
2) replace t.ev slice with a linked list and track "seen" using pointers, relying on GC to reclaim the memory.

(1) looks relatively simple and efficient but implies certain added coordination (locking), making it more challenging to get right;
(2) would be much-much simpler implementation-wise but puts an extra strain on the standard GC.

I'm in progress of prototyping (1) but it's already clear it's quite a piece on its own. Let me know what you think, @anacrolix.

anacrolix:
Over several days with Torrents that have seen a lot of connection churn, the memory use grows to be probably the largest idle consumer of memory space and object count. The other biggest consumers are connection read buffers (which is reasonable for active connections), and piece state tracking (very large torrents, separate problem).

The main reason I haven't seen this earlier is I haven't tested with seeding enabled, and without dropping Torrents at regular intervals.

1. Sounds quite good, but 2 sounds pretty neat. I guess you can't use list.List as it's doubly-linked? Would you just do a custom singly-listed implementation?

anacrolix:
Happy for you to explore some ideas, I can run them in my production scenario for comparison if you like. I think it's worth fixing this issue, but it's not urgent to get something out.

yarikk:
I ended up abandoning (1) in favour of (2). #693 implements that.

anacrolix:
Here's a before, with `focus=pex`

![image](https://user-images.githubusercontent.com/988750/145518069-b8c6b70d-7cdf-4f2b-a158-2c2501adb984.png)

[...]

PR #692, opened 2021-12-06
PEX: manage memory consumed by event log

yarikk:
As discussed under #691 this implements the "_shift-and-truncate_" approach (1):
> 1. periodically shift-and-truncate `t.pex.ev` based on minimal values for `t.conn[*].pex.seq`, tracking the virtual offset after truncations

This is an early draft: it is nowhere near ready yet. I'm putting it out to attract an early feedback.


anacrolix:
It's a nice start, where do you think it should be called from?

anacrolix:
I'm fixing the CI issue separately, it's unrelated.

yarikk:
Abandoning this in favour of #693.

PR #700, opened 2021-12-23
http trackers query string kept at original positon

boypt:
Some http private tracker requires the original URL querystring (contains a unique user key) to be in the first position. this is not following the HTTP protocol but understandable for anti-DDOS purposes.

The original logic is to add other arguments to the Query object, then re-encoded to be the query string, this cause the key position changed.

This PR make sure that the original query param kept in the original position if appears.


anacrolix:
Thanks!

PR #718, opened 2022-01-31
test: use `T.TempDir` to create temporary test directory

Juneezee:
We can write less code by using the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete.

Reference: https://pkg.go.dev/testing#T.TempDir


anacrolix:
Thanks. CI was broken, hopefully now it should be fixed.

PR #722, opened 2022-02-04
Pass client logger to anacrolix/go-libutp sockets

### LINKED ISSUE #676: ClientConfig Logger should be used for utp
Per https://github.com/anacrolix/torrent/issues/332#issuecomment-927179066, the ClientConfig Logger should be passed through to go-libutp (and possibly anacrolix/utp when that's in use?). Currently it's necessary to modify the global Logger instance manually, and that belongs to the package, not to any utp instances owned by a torrent Client.

FIGBERT:
I'm opening this as a draft for now, because I'm not certain I like the use of `utp.NewSocketOpt` all the way up the chain in `listenAll`.

I'd love your input, @anacrolix, but I was thinking of perhaps defining a `socket` interface in addition to the `utpSocket` interface in `utp.go`, and then a `socketOpt` type which could be used for configuring sockets more generally.

Closes #676.

### DIFF
--- a/client.go
+++ b/client.go
@@ -20,6 +20,7 @@ import (
 	"github.com/anacrolix/chansync/events"
 	"github.com/anacrolix/dht/v2"
 	"github.com/anacrolix/dht/v2/krpc"
+	utp "github.com/anacrolix/go-libutp"
 	"github.com/anacrolix/log"
 	"github.com/anacrolix/missinggo/perf"
 	"github.com/anacrolix/missinggo/pubsub"
@@ -246,7 +247,7 @@ func NewClient(cfg *ClientConfig) (cl *Client, err error) {
 		}
 	}
 
-	sockets, err := listenAll(cl.listenNetworks(), cl.config.ListenHost, cl.config.ListenPort, cl.firewallCallback)
+	sockets, err := listenAll(cl.listenNetworks(), cl.config.ListenHost, cl.config.ListenPort, cl.firewallCallback, utp.WithLogger(cl.logger))
 	if err != nil {
 		return
 	}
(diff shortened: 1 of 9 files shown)


anacrolix:
Thank you for trying this. Note the CI fails as the CGO_ENABLED=0 case is not handled. What about instead of passing the option all the way down, you pass the logger. Then at the utp/go-libutp build separation point, you use the logger appropriately for each package there? That would be the easier solution. A more difficult one might be to pass an option type generic to each of those packages and defined based on the build tags. However anacrolix/utp doesn't support a custom logger as far as I know, and it's deprecated so it's probably not worth the effort. Cheers!

FIGBERT:
I've updated the PR to pass the logger instead of the option. The `log.Logger` arg is required now, which we could play with, but I'm not certain how to make it optional while preserving the idea that it's singular (as opposed to `...log.Logger`).

anacrolix:
I believe it's all unexported code inside anacrolix/torrent that deals with this. So just pass the entire (singular Logger), always.

anacrolix:
Apologies, that's now triggered.

anacrolix:
If you merge master in to your branch, you can run against the improved CI. Or rebase your branch on to master. Let me know.

FIGBERT:
Rebased my branch against `master` for the improved CI, so you'll need to approve the workflow again. Thanks!

FIGBERT:
Aha double rebase

FIGBERT:
I've now marked the PR as ready for review, and thus it's ~~probably~~ ready for merging into the `master` branch (and the release of a new patch[?] version).

anacrolix:
Great work!

PR #734, opened 2022-04-01
Avoid heap allocation in GetRequestablePieces

u1f35c:
The calculation of whether we should ignore a piece in GetRequestablePieces ends up doing an allocation for every piece, when
all we really need to do is query the index in the torrent. Provide an IgnorePiece function instead, which avoids the need for a temporary allocation.

Observed to cut out 40% of object allocations in some workloads (large download, lots of seeds).

(I'm not clear why Go doesn't optimise this onto the stack, as it doesn't seem to escape, but it was turning up as an extremely hot allocation path in pprof flame graphs.)

### DIFF
--- a/request-strategy-impls.go
+++ b/request-strategy-impls.go
@@ -44,8 +44,15 @@ type requestStrategyTorrent struct {
 	t *Torrent
 }
 
-func (r requestStrategyTorrent) Piece(i int) request_strategy.Piece {
-	return requestStrategyPiece{r.t, i}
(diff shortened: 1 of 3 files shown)


anacrolix:
Good catch. This oddity might be a hold out from when there were actually other implementations of request_strategy.Piece. I'll wait for CI.

anacrolix:
I took a look at keeping the Piece interface, by allocating it somewhere on Torrent, but it wasn't very convincing. Let's try this.

Issue #720, opened 2022-02-02
logs without level about "retransmission failure"


AskAlexSharov:
version 1.40.0
```
sctp ERROR: 2022/02/02 15:41:22 [0xc102917c00] retransmission failure: T1-init
sctp ERROR: 2022/02/02 15:41:22 [0xc141a968c0] retransmission failure: T1-cookie
sctp ERROR: 2022/02/02 15:41:22 [0xc05921e380] retransmission failure: T1-init
```

anacrolix:
Good find. What do you think about me implementing the webrtc logging interface to tap into anacrolix/torrent.Client's logger? Looks like they've exposed it quite nicely. I wonder if this error in general is something that needs to be reported upstream.

anacrolix:
https://gophers.slack.com/archives/CAK2124AG/p1643854599797609

AskAlexSharov:
Good for me, thank you.

anacrolix:
Fixed by #735.

PR #735, opened 2022-04-09
Discard webrtc logs

AskAlexSharov:
Problem:
- Now user can see logs like: `mux ERROR: 2022/04/09 13:41:02 mux: ending readLoop dispatch error io: read/write on closed pipe` (when restart the app), but can't do anything with this logs (filter/format).

Probably it's bad solution (to discard logs), but It's unclear for me - how to integrate pion logger with anacrolix logger: 
- api *webrtc.API - is the global variable initialized way before i creating configuration of anacrolix

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -60,7 +60,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: ./.github/actions/go-common
     - name: Some packages compile for WebAssembly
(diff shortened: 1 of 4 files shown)


anacrolix:
I think this is related to https://github.com/anacrolix/torrent/issues/720, is that right @AskAlexSharov ?

AskAlexSharov:
yes

PR #737, opened 2022-04-21
[webseed] Add a custom URL encoder for webseeds

afjoseph:
Heyo! Thanks for the great project!

This PR abstracts some of the webseed logic to allow customizing URL encodes. Currently, it's hardcoded to use `url.QueryEscape`. Some providers use different escaping mechanisms (e.g., `url.PathEscape`).

After this PR is merged, one can use it thusly:

```
	spec := &torrent.TorrentSpec{
		Trackers:    [][]string{...},
		InfoHash:    ...,
		DisplayName: ...,
		EncodeWebSeedUrl: func(u string) string {
			if strings.Contains(u, "blahblah") {
				return url.PathEscape(url)
			} else {
				return url.QueryEscape(url)
			}
		},
	}
	cl := torrent.NewClient(...)
	cl.AddTorrentSpec(spec)
```

### DIFF
--- a/spec.go
+++ b/spec.go
@@ -36,6 +36,10 @@ type TorrentSpec struct {
 	// Whether to allow data download or upload
 	DisallowDataUpload   bool
 	DisallowDataDownload bool
+
+	// Custom encoder for webseed URLs
+	// Leave nil to use the default (url.QueryEscape)
+	EncodeWebSeedUrl func(string) string
 }
 
 func TorrentSpecFromMagnetUri(uri string) (spec *TorrentSpec, err error) {
--- a/torrent.go
+++ b/torrent.go
@@ -2350,15 +2350,28 @@ func (t *Torrent) callbacks() *Callbacks {
 	return &t.cl.config.Callbacks
 }
 
-func (t *Torrent) AddWebSeeds(urls []string) {
+type AddWebSeedsOptions struct {
+	// Custom encoder for webseed URLs
+	// Leave nil to use the default (url.QueryEscape)
+	EncodeUrl func(string) string
+}
+
+
+// Add web seeds to the torrent.
+// If AddWebSeedsOptions is not nil, only the first one is used.
+func (t *Torrent) AddWebSeeds(urls []string, opts ...AddWebSeedsOptions) {
 	t.cl.lock()
 	defer t.cl.unlock()
 	for _, u := range urls {
-		t.addWebSeed(u)
+		if opts == nil {
+			t.addWebSeed(u, nil)
+		} else {
+			t.addWebSeed(u, opts[0].EncodeUrl)
+		}
 	}
 }
 
-func (t *Torrent) addWebSeed(url string) {
+func (t *Torrent) addWebSeed(url string, encodeUrl func(string) string) {
 	if t.cl.config.DisableWebseeds {
 		return
 	}
@@ -2395,6 +2408,7 @@ func (t *Torrent) addWebSeed(url string) {
 					r: r,
 				}
 			},
+			EncodeUrl: encodeUrl,
 		},
 		activeRequests: make(map[Request]webseed.Request, maxRequests),
 		maxRequests:    maxRequests,
--- a/webseed/request.go
+++ b/webseed/request.go
@@ -12,28 +12,56 @@ import (
 
 // Escapes path name components suitable for appending to a webseed URL. This works for converting
 // S3 object keys to URLs too.
+// Contrary to the name, this actually does a QueryEscape, rather than a
+// PathEscape. This works better with most S3 providers. You can use
+// EscapePathWithCustomEncoding for a custom encoding.
 func EscapePath(pathComps []string) string {
+	return escapePath(pathComps, nil)
+}
+
+func escapePath(pathComps []string, encodeUrl func(string) string) string {
+	if encodeUrl == nil {
+		encodeUrl = url.QueryEscape
+	}
 	return path.Join(
 		func() (ret []string) {
 			for _, comp := range pathComps {
-				ret = append(ret, url.QueryEscape(comp))
+				ret = append(ret, encodeUrl(comp))
 			}
 			return
 		}()...,
 	)
 }
 
-func trailingPath(infoName string, fileComps []string) string {
-	return EscapePath(append([]string{infoName}, fileComps...))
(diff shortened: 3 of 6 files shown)


spec.go
```diff
@@ -36,6 +36,10 @@ type TorrentSpec struct {
 	// Whether to allow data download or upload
 	DisallowDataUpload   bool
 	DisallowDataDownload bool
+
+	// Custom encoder for webseed URLs
+	// Leave nil to use the default (url.QueryEscape)
+	EncodeWebSeedUrl func(string) string
```
anacrolix, line 42: How about `DefaultWebseedEscapePath`, where default here means the escaping that is used when webseed URLs are given through API constructs that don't take an option to specify an alternative escaping?

torrent.go
```diff
@@ -2350,15 +2350,28 @@ func (t *Torrent) callbacks() *Callbacks {
 	return &t.cl.config.Callbacks
 }
 
-func (t *Torrent) AddWebSeeds(urls []string) {
+type AddWebSeedsOptions struct {
```
anacrolix, line 2353: I don't see any example use of this in this PR, but maybe using the function-opt pattern is more forward compatible here: `type WebseedOpt func(*webseedOpts)`, and provide `PathEscaper(pathComps []string) string` as the only instance of it for now. Then AddWebSeeds takes `...WebseedOpt` or `...AddWebseedOpt`.

anacrolix (commented):
I think I would define a `type PathEscaper(pathComps []string) string` which is the widest possible signature for the escaping (it will handle all possible variants on this). It's always `pathComps []string` in, and the output is always added to the end of the base URL that ended with `/`, that never changes. What comes after could be anything and this signature captures it.

I would also refer to it as "Escape Path" or whatever variants on that throughout. The nomenclature in Go is escape, and we're only ever doing this to the path (both torrent path and URL path in this instance).

webseed/request_test.go
```diff
@@ -23,7 +23,7 @@ func TestTrailingPath(t *testing.T) {
 }
 
 func TestTrailingPathForEmptyInfoName(t *testing.T) {
-	qt.Check(t, trailingPath("", []string{`ノ┬─┬ノ ︵ ( \o°o)\`}), qt.Equals, "%E3%83%8E%E2%94%AC%E2%94%80%E2%94%AC%E3%83%8E+%EF%B8%B5+%28+%5Co%C2%B0o%29%5C")
-	qt.Check(t, trailingPath("", []string{"hello", "world"}), qt.Equals, "hello/world")
-	qt.Check(t, trailingPath("war", []string{"and", "peace"}), qt.Equals, "war/and/peace")
+	qt.Check(t, trailingPath("", []string{`ノ┬─┬ノ ︵ ( \o°o)\`}, url.QueryEscape), qt.Equals, "%E3%83%8E%E2%94%AC%E2%94%80%E2%94%AC%E3%83%8E+%EF%B8%B5+%28+%5Co%C2%B0o%29%5C")
+	qt.Check(t, trailingPath("", []string{"hello", "world"}, url.QueryEscape), qt.Equals, "hello/world")
+	qt.Check(t, trailingPath("war", []string{"and", "peace"}, url.QueryEscape), qt.Equals, "war/and/peace")
 }
```
anacrolix, line 29: Another test for example for an encoding that handles commas or spaces differently could be useful here, at least just to show that the default query escaping isn't the only one that works.

webseed/request.go
```diff
@@ -12,28 +12,56 @@ import (
...
+func trailingPath(infoName string, fileComps []string, encodeUrl func(string) string) string {
+	return escapePath(append([]string{infoName}, fileComps...), encodeUrl)
 }
 
 // Creates a request per BEP 19.
 func NewRequest(url_ string, fileIndex int, info *metainfo.Info, offset, length int64) (*http.Request, error) {
+	return newRequest(url_, fileIndex, info, offset, length, nil)
+}
+
+func NewRequestWithCustomUrlEncoding(
```
anacrolix, line 45: What about `NewRequestWithOpts` and a `NewRequestOpt` struct that contains `EscapePath PathEscaper`?

torrent.go
afjoseph: Done https://github.com/afjoseph/torrent/commit/a79c3bd5d30d99fb33ab7c5334c7d017884d4403

webseed/request_test.go
afjoseph: Done https://github.com/afjoseph/torrent/commit/a79c3bd5d30d99fb33ab7c5334c7d017884d4403

webseed/request.go
afjoseph: Done https://github.com/afjoseph/torrent/commit/a79c3bd5d30d99fb33ab7c5334c7d017884d4403

spec.go
afjoseph: Done https://github.com/afjoseph/torrent/commit/a79c3bd5d30d99fb33ab7c5334c7d017884d4403

anacrolix: I think it's actually better to avoid the extra option just for webseeds entirely (since each URL can require different options).

torrent.go
anacrolix: I've threaded this through completely in a fix up.

webseed/request.go
anacrolix: I know I mentioned that NewRequest was public API, but at the cost of enumerating all the option styles for creating a Request, and not being able to find anywhere that it's actually used publicly, I've made it private instead.

anacrolix:
> I made an extra commit [here](https://github.com/anacrolix/torrent/pull/737/commits/7571864ec6c9e9f806dbba2248d624f327e1df79) to address an [explicit panic](https://github.com/afjoseph/torrent/blob/7571864ec6c9e9f806dbba2248d624f327e1df79/client.go#L1151) that occurs when IP is nil. It happened quite a bit of times recently for whatever reason. I added it alongside the port check and covered it with a regression test

I think someone else also reported this (and may have fixed it), I'll follow it up.

anacrolix:
@afjoseph could you provide the stack trace that gave you the panic in `lient.badPeerIPPort`?

anacrolix:
See https://github.com/anacrolix/torrent/tree/afjoseph which should include everything mentioned above. If you're happy with that, let's close this PR.

afjoseph:
Very happy with it and many thanks. Closing the PR

PR #740, opened 2022-04-23
Use Torrent logger instead of default logger

FIGBERT:
The crusade continues!


ramezanius (approved):

anacrolix:
Thank you! And @ramezanius too

PR #743, opened 2022-05-05
Make pp.Decoder.MaxLength configurable

AskAlexSharov:
For https://github.com/anacrolix/torrent/discussions/741

### DIFF
--- a/config.go
+++ b/config.go
@@ -10,7 +10,6 @@ import (
 	"github.com/anacrolix/dht/v2/krpc"
 	"github.com/anacrolix/log"
 	"github.com/anacrolix/missinggo/v2"
-	pp "github.com/anacrolix/torrent/peer_protocol"
 	"github.com/anacrolix/torrent/version"
 	"golang.org/x/time/rate"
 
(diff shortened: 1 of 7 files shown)


config.go
```diff
@@ -155,6 +156,10 @@ type ClientConfig struct {
 	DisableWebtorrent bool
 	DisableWebseeds   bool
 
+	// Drop any peer response bigger than this. To prevent DoS or abuse.
+	// TorrentSpec.ChunkSize can't be bigger than this value
+	MaxResponseLength pp.Integer
```
anacrolix, line 161: How about `DecodePeerMessageMaxLength`?

anacrolix: Or `PeerMessageDecoderMaxLength`

AskAlexSharov:
renamed to PeerMessageDecoderMaxLength, fixed tests

anacrolix:
Very nice! Just before I squash and merge it @AskAlexSharov : any reason not to just set the max decoder length based on the Torrent chunk size? (say max(4*chunkSize, ~24KiB) or so?

AskAlexSharov:
@anacrolix ok, changed to `4 * max(t.chunkSize, defaultChunkSize)` and removed global config

anacrolix:
Thanks, it's probably simpler not to expose the config option and infer it automatically like you've done now. I think the config is already complex enough. I'll merge this when the tests pass.

PR #750, opened 2022-05-18
Expose `Peer.DownloadRate`

quantumsheep:
This function is very useful and could be accessible to library users.


anacrolix:
Looks good, I'll probably merge after CI runs.

PR #755, opened 2022-06-10
Do t.storage.Flush() on torrent completion, and on storage.Close() 

AskAlexSharov:
Only question: t.updateComplete is inside lock, is it ok to call syncronious Flush (go-mmap lib doesn't have method for `unix.MS_ASYNC`) there?


mmap_span/mmap_span.go
```diff
@@ -19,11 +19,30 @@ func (ms *MMapSpan) Append(mMap mmap.MMap) {
...
+	ms.mu.Lock()
+	defer ms.mu.Unlock()
+	for _, mMap := range ms.mMaps {
+		err := mMap.Flush()
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+	// This is for issue 211.
+	ms.mMaps = nil
```
anacrolix, line 32: This seems strange, why woudl Flush do this, especially if htere is no error?

torrent.go
```diff
@@ -2461,7 +2461,11 @@ func (t *Torrent) pieceRequestIndexOffset(piece pieceIndex) RequestIndex {
 }
 
 func (t *Torrent) updateComplete() {
+	before := t.Complete.Bool()
```
anacrolix, line 2464: I think there might be a more suitable place to put this, there should be a method that's only called when torrent completion transitions to complete from incomplete.

bad_storage.go
```diff
@@ -17,6 +17,7 @@ var _ storage.ClientImpl = badStorage{}
 func (bs badStorage) OpenTorrent(*metainfo.Info, metainfo.Hash) (storage.TorrentImpl, error) {
 	return storage.TorrentImpl{
 		Piece: bs.Piece,
+		Flush: func() error { return nil },
```
anacrolix, line 20: I think it's common enough not to have a Flush method, that teh caller can check for Flush != nil in the TorrentImpl struct.

anacrolix:
I think maybe the asynchronous form should be preferred, unless the client lock is not held. I can take another more careful look when the PR has progressed some more.

torrent.go
AskAlexSharov: sorry, I don't see such method.

bad_storage.go
AskAlexSharov: done

mmap_span/mmap_span.go
AskAlexSharov: yes, mistake. removed.

torrent.go
anacrolix: Ah I see the confusion. This flushes the entire torrent, when the entire torrent is complete. Do you want perhaps flush the individual memory-maps, when associated pieces become complete? That would be a more gradual approach.

anacrolix: Sorry for not spotting this before. Similarly, if this was the approach, the Flush would move on to the piece implementation. Let me know your thoughts.

AskAlexSharov:
done

anacrolix:
Looks like this PR is turning out to be a bit of a struggle. I took a good look, I can't push to your branch without you letting collaborators edit, but I think I found where you want the flush to occur (for piece completion). See https://github.com/anacrolix/torrent/commit/2c56fa7acb915dd81e0b4903bd13b50c2d7ee479. Let's address that first, and then do the sync on Close later when it is simpler. I notice this PR currently doesn't use the flush at all, that needs to be linked back up via the Piece storage wrapper.

AskAlexSharov:
done

AskAlexSharov:
@anacrolix race between: 
`if p.hasDirtyChunks() {`
and
`piece.unpendChunkIndex(`
I don't see any existing mutex in piece
and probably move `if p.hasDirtyChunks() {` before `t.cl.unlock()` will not help

how better handle it?

anacrolix:
It was just demonstration code for where to put it. You probably do want to check for dirty pieces inside the locked region, then pass out a boolean for it to do the flushing.

AskAlexSharov:
CI green

PR #760, opened 2022-06-25
Add ability to set DialContext/ListenPacket for tracker announcements

ccampbell:
I have had this branch lying around for a while, but decided to pick it back up, make it work with the latest master, and offer it as a PR.

Currently, this library has the ability to configure an HTTPProxy for HTTP tracker announcements. This is useful for a simple proxy, but it is somewhat limited if you want to do anything more advanced. Likewise, there is currently no way to proxy UDP tracker announcements.

This PR introduces two new configuration properties for `TrackerDialContext` and `TrackerListenPacket`. When they are defined they will take precedence over the default `http.Transport` options and `net.ListenPacket`.

Typically these would not be required since most people probably use a VPN at a lower level to prevent IP leaks, but these changes allow you to, for example, route all tracker requests through a remote wireguard server even if your computer is not connected to a VPN.

### DIFF
--- a/config.go
+++ b/config.go
@@ -1,6 +1,7 @@
 package torrent
 
 import (
+	"context"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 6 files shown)


anacrolix:
@ccampbell Thanks for contributing. It looks like a solid PR. Give me a day or two to think it over, cheers.

anacrolix:
Looks good. I was worried that it conflicted with the DNS checking options that were added recently by @afjoseph but I don't think that's the case.

PR #761, opened 2022-07-03
Reproduce "unlock of unlocked mutex"

AskAlexSharov:
see tests results


anacrolix:
Panicking is not expected in that location. If you need to recover more gracefully from panics there, use another function and defer the locks/unlocks as appropriate.

PR #766, opened 2022-07-13
use RLock where can

AskAlexSharov:


client.go
```diff
@@ -1374,8 +1374,8 @@ func (cl *Client) allTorrentsCompleted() bool {
 // Returns true when all torrents are completely downloaded and false if the
 // client is stopped before that.
 func (cl *Client) WaitAll() bool {
-	cl.lock()
-	defer cl.unlock()
+	cl.rLock()
```
anacrolix: I don't think you can RLock if you plan to use Cond.Wait, as it wants to Unlock. The other changes look good though.

anacrolix (changes requested):

anacrolix:
Saw your update, should be fine to merge when the tests pass 👍🏻 .

client.go
AskAlexSharov: yes. fixed

PR #773, opened 2022-08-14
Use consistent receiver names across methods.

veshij:
There are a few places in the codebase where different methods of the same object have different receiver names.
Sometimes it might be slightly confusing, updating the code to have consistent naming.


anacrolix:
I'm not sure this is worth the churn. It segments the history in a lot of places on a non-functional change. I think Go messed up by not having an implicit self/this/me, or at least not enforcing a better convention on it.

Issue #784, opened 2022-11-10
Adding custom authentication headers


marcovidonis:
I'm looking at adding custom headers to http or websocket requests that I send from the client to a tracker I control. The idea is to filter the requests coming to the tracker to only allow the ones from clients authenticated via a token.

It doesn't seem like there's currently an option for this in the client. How would you approach this @anacrolix?

anacrolix:
Hm, I suspect something like the [`Director`](https://pkg.go.dev/net/http/httputil@go1.19.3#ReverseProxy) field should be exposed on `torrent.ClientConfig`. It should be a func that takes and modifies the `http.Request` before it's sent to the tracker. It would probably run just after the user agent and host header fields are handled, [here](https://github.com/anacrolix/torrent/blob/30a5d079be25dda0c69874ee09cb87b8d53cc52a/tracker/http/http.go#L89-L98). Thanks!

marcovidonis:
Thanks for the feedback, I've followed your suggestions and made it so that we can apply a user-defined function that modifies the `http.Request` just before it's sent. I could open a PR for it if you'd like.

My question now is, the way I've set it up, it's not limited to custom headers, as I simply run any function that modifies the request. Is this an acceptable behaviour, or should I put some limits in place, for example just allow adding new headers?

marcovidonis:
I actually just realised that #787 didn't cover the case where the connection to the tracker is done via a websocket, as `HttpRequestDirector` is not applied to the `Upgrade` request.

I'm thinking of exposing `HttpRequestDirector` in `webtorrent.TrackerClient` and calling it just before `Dial()`, [here](https://github.com/anacrolix/torrent/blob/master/webtorrent/tracker-client.go#L89).
However, the HTTP request is not available in `TrackerClient`. The Dialer takes a `header` as an optional parameter, so I thought I could just create an empty `http.Request`, apply `HttpRequestDirector` and then extract the header from the request, to be used in the Dialer.

 I'm not 100% sure of this approach, though. On one hand, it feels a bit hacky, but on the other, I prefer using the same Director function rather than making another method in `ClientConfig` just to add headers. Any thoughts @anacrolix?

anacrolix:
[...]

PR #787, opened 2022-11-28
Modify HTTP request before sending

marcovidonis:
Exposes a method in `torrent.ClientConfig` to define a function that modifies the HTTP request. The method is called right before sending the request.

This could be used, for example, to add authentication headers to the request, which could be used to grant permission to a private tracker.

This addresses #784.

### DIFF
--- a/tracker/http/http.go
+++ b/tracker/http/http.go
@@ -99,7 +99,7 @@ func (cl Client) Announce(ctx context.Context, ar AnnounceRequest, opt AnnounceO
 
 	err = opt.HTTPRequestDirector(req)
 	if err != nil {
-		err = fmt.Errorf("error applying custom HTTP request headers: %s", err)
+		err = fmt.Errorf("error modifying HTTP request: %s", err)
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


tracker/http/http.go
```diff
@@ -97,10 +97,12 @@ func (cl Client) Announce(ctx context.Context, ar AnnounceRequest, opt AnnounceO
...
 	}
 
-	err = opt.HTTPRequestDirector(req)
-	if err != nil {
-		err = fmt.Errorf("error modifying HTTP request: %s", err)
-		return
+	if opt.HTTPRequestDirector != nil {
+		err = opt.HTTPRequestDirector(req)
+		if err != nil {
+			err = fmt.Errorf("error modifying HTTP request: %s", err)
```
anacrolix, line 103:
```suggestion
			err = fmt.Errorf("error modifying HTTP request: %w", err)
```

anacrolix:
I might expedite this and merge it in. I also will make some opinionated changes to some config field names, I believe the Go team got the convention wrong, and it's starting to spread inside the code base. The failing CI is unrelated and fixed on master.

PR #789, opened 2022-12-05
Add customer headers when dialling WS connection to tracker

marcovidonis:
Expose a `WebtorrentTrackerHttpHeader` field in Client.config to let users define a function that returns an `http.Header` to be used in `Dialer.Dial()` when establishing a connection. Useful for adding authentication headers when using a private tracker, for example.

Addresses #784.

### DIFF
--- a/client.go
+++ b/client.go
@@ -297,9 +297,9 @@ func NewClient(cfg *ClientConfig) (cl *Client, err error) {
 			}
 			return t.announceRequest(event), nil
 		},
-		Proxy:                       cl.config.HTTPProxy,
-		WebtorrentTrackerHttpHeader: cl.config.WebtorrentTrackerHttpHeader,
-		DialContext:                 cl.config.TrackerDialContext,
+		Proxy:                      cl.config.HTTPProxy,
+		WebsocketTrackerHttpHeader: cl.config.WebsocketTrackerHttpHeader,
+		DialContext:                cl.config.TrackerDialContext,
 		OnConn: func(dc datachannel.ReadWriteCloser, dcc webtorrent.DataChannelContext) {
 			cl.lock()
 			defer cl.unlock()
--- a/client.go
+++ b/client.go
@@ -297,8 +297,9 @@ func NewClient(cfg *ClientConfig) (cl *Client, err error) {
 			}
 			return t.announceRequest(event), nil
 		},
-		Proxy:       cl.config.HTTPProxy,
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -105,6 +105,9 @@ type ClientConfig struct {
 	// HttpRequestDirector modifies the request before it's sent.
 	// Useful for adding authentication headers, for example
 	HttpRequestDirector func(*http.Request) error
+	// WebtorrentTrackerHttpHeader returns a custom header to be used when dialing a websocket connection
+	// to the tracker. Useful for adding authentication headers
+	WebtorrentTrackerHttpHeader func() (http.Header, error)
```
anacrolix, line 110: Is it possible to do this without the error return?

client.go
```diff
@@ -297,8 +297,9 @@ func NewClient(cfg *ClientConfig) (cl *Client, err error) {
 			}
 			return t.announceRequest(event), nil
 		},
-		Proxy:       cl.config.HTTPProxy,
-		DialContext: cl.config.TrackerDialContext,
+		Proxy:                       cl.config.HTTPProxy,
+		WebtorrentTrackerHttpHeader: cl.config.WebtorrentTrackerHttpHeader,
```
anacrolix, line 301: What made you choose Webtorrent over Websocket? I guess I named it webtorrent in the repo, but websocket without the tracker context isn't very specific. Just wonder if you have any thoughts either way.

config.go
marcovidonis: Yeah sure! So a function that just returns an `http.Header` then?

client.go
marcovidonis: I was a bit unsure about which to choose, to be honest. I think Webtorrent might make it easier to understand what the purpose of setting this header is (communicating with a webtorrent tracker) when defining the client config. Also, I thought it would be consistent with the way the project is organised, with all websocket related parts collected under webtorrent (as far as I can tell). 
However, I realise that it might be more consistent if a type named `websocketTrackers` had a field called `WebsocketTrackerHttpHeader`... Or we might simply name it `HttpHeader` in `websocketTrackers` and something more descriptive in `Client.config`.

All in all, I don't really have a strong preference either way, whatever will make it easier to understand what the function is supposed to do :)

config.go
anacrolix: Yeah

anacrolix:
Fantastic, thanks!

PR #796, opened 2022-12-23
Change default webseed path escaping to work for all S3-compatible pr…

anacrolix:
…oviders

This seems to work regardless of the S3 provider.


anacrolix:
@afjoseph did you want to take a look at this, it follows up from your work earlier.

PR #803, opened 2023-01-20
Add WebRTC peer connection transport stats

marcovidonis:
Looks like there is still some timing issue, whereby sometimes `tc.transportStats` may be `null` if the file transfer is too short (small file on same machine, for example).
For larger file transfers, there is a high variability in the number of peer connections being created during the transfer (i.e. `len(tc.transportStats)`). Is this real or again a timing issue when appending new stats to the slice?

Still in draft for now.

### DIFF
--- a/webtorrent/tracker-client.go
+++ b/webtorrent/tracker-client.go
@@ -43,6 +43,8 @@ type TrackerClient struct {
 	pingTicker     *time.Ticker
 
 	WebsocketTrackerHttpHeader func() http.Header
+
+	transportStats []webrtc.StatsReport
 }
 
 func (me *TrackerClient) Stats() TrackerClientStats {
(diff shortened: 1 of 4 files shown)


webtorrent/tracker-client.go
```diff
@@ -233,6 +235,14 @@ func (tc *TrackerClient) Announce(event tracker.AnnounceEvent, infoHash [20]byte
 		return fmt.Errorf("creating offer: %w", err)
 	}
 
+	pc.OnClose(func() {
+		stats := pc.GetStats()
+		if tc.transportStats == nil {
```
anacrolix, line 240: You can append to nil slices, no need for this check!

webtorrent/transport.go
```diff
@@ -39,16 +39,35 @@ type wrappedPeerConnection struct {
...
 	ctx  context.Context
+
+	onCloseHandler func()
 }
 
 func (me *wrappedPeerConnection) Close() error {
 	me.closeMu.Lock()
 	defer me.closeMu.Unlock()
+
+	me.onClose()
```
anacrolix, line 50: Do you need to guard against multiple closes here?

anacrolix:
Is it possible that connections aren't being closed, and that's why you're not seeing stats? Re the leak I mentioned.

webtorrent/transport.go
marcovidonis: I'm not super confident with locks, but my understanding is that here `me.onClose()` is called while `me` is still locked, right? I've noticed that if I call it outside the lock, we end up trying to access the stats after the connection is already closed, so we only get `nil` values. But I'm open to suggestions!

anacrolix: Ah no I actually mean repeat closes. Just a sync.Once to wrap it perhaps.

PR #812, opened 2023-02-11
Remove unecessary completion "set"

colinmarc:
If the file has never been downloaded, complete will naturally be false. It's
not necessary to then set it false again unless it was actually claimed to be
true in the first place.

In my tests, using the boltdb completion thingy with fsync turned *on*, this
reduced the cold start for big buck bunny from multiple seconds to just a few
ms.

### DIFF
--- a/storage/file-piece.go
+++ b/storage/file-piece.go
@@ -28,20 +28,25 @@ func (fs *filePieceImpl) Completion() Completion {
 		c.Ok = false
 		return c
 	}
+
+	verified := true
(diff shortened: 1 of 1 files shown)


anacrolix:
This is good, but does it have the unintended side affect of no longer setting `Complete` to false in the return value if verified is false?

colinmarc:
Yes it did, fixed now!

anacrolix:
Thanks!

PR #823, opened 2023-03-25
fix panic in infohash.FromHexString

azr:
Howdy ! I managed to provoke this one somehow, so here's a fix.

### DIFF
--- a/types/infohash/infohash.go
+++ b/types/infohash/infohash.go
@@ -46,7 +46,8 @@ func (t *T) FromHexString(s string) (err error) {
 		return
 	}
 	if n != Size {
-		panic(n)
+		err = fmt.Errorf("decoded hash hex string has bad length: %d", n)
+		return
 	}
 	return
 }


types/infohash/infohash.go
```diff
@@ -46,7 +46,8 @@ func (t *T) FromHexString(s string) (err error) {
 		return
 	}
 	if n != Size {
-		panic(n)
+		err = fmt.Errorf("decoded hash hex string has bad length: %d", n)
```
azr, line 49: probably initially gave the right length, but the string was garbage.

anacrolix: @azr what input string was able to cause the panic? I added a fuzz test to try to find anything and I'm not having any luck.

anacrolix:
@azr did you have any insight into the comment I made?

types/infohash/infohash.go
anacrolix: Bummer. I'll leave it as is, if you encounter the problem again, let me know.

PR #824, opened 2023-03-31
Add WebRTC ICE servers config

marcovidonis:
This PR adds an optional field to the Client config, where the user can specify a list of ICE Server URLs that will be passed to Pion as part of the WebRTC config when creating a new WebRTC Peer Connection.

There is already an option for this in Pion, this PR exposes it to the torrent client:
>`ICEServers` defines a slice describing servers available to be used by ICE, such as STUN and TURN servers.


webtorrent/transport.go
```diff
@@ -49,11 +49,17 @@ func (me *wrappedPeerConnection) Close() error {
...
 
-func newPeerConnection(logger log.Logger) (*wrappedPeerConnection, error) {
+func newPeerConnection(logger log.Logger, iceServers []string) (*wrappedPeerConnection, error) {
 	newPeerConnectionMu.Lock()
 	defer newPeerConnectionMu.Unlock()
 	ctx, span := otel.Tracer(tracerName).Start(context.Background(), "PeerConnection")
-	pc, err := api.NewPeerConnection(config)
+
+	pcConfig := config
+	if iceServers != nil {
```
anacrolix, line 58: Is this check necessary? Can you just have a default value in the ClientConfig, or will pion treat nil as the default anyway?

anacrolix:
@marcovidonis just following up, do you still want to get this in?

marcovidonis:
Sorry for the long absence @anacrolix! Yes, I plan on picking this up next week it that's alright.

webtorrent/transport.go
marcovidonis: This was in fact an unnecessary check. If Pion receives `nil` or an empty list of ICE Servers, it will simply ignore it.

anacrolix:
Cheers

PR #835, opened 2023-05-17
downloader: speedup "--downloader.verify" mode

AskAlexSharov:
deduplicate logic
create more producer goroutines (torrent lib does limiting internally amount of consumers/disk-readers/hashers by 2, and it's enough because we can verify multiple files in parallel)
move flag from "downloader torrent_hashes --verify" to "downloader --verify"


AskAlexSharov:
oops. wrong repo.

PR #838, opened 2023-05-24
Reintroduce torrent-wide PEX throttling

anacrolix:
This is what I'm thinking if https://github.com/anacrolix/torrent/discussions/836#discussioncomment-5984103 is correct?

### DIFF
--- a/pex.go
+++ b/pex.go
@@ -4,6 +4,7 @@ import (
 	"net"
 	"net/netip"
 	"sync"
+	"time"
 
 	pp "github.com/anacrolix/torrent/peer_protocol"
 )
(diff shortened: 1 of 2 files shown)


pex.go
```diff
@@ -144,8 +145,11 @@ func (me *pexMsgFactory) PexMsg() *pp.PexMsg {
 // Per-torrent PEX state
 type pexState struct {
 	sync.RWMutex
-	tail *pexEvent     // event feed list
-	hold []pexEvent    // delayed drops
+	tail *pexEvent  // event feed list
```
yarikk, line 148: This one here, btw, is not related to throttling: it's needed to send out the follow-up deltas every minute, as provisioned in the BEP. It's good to recover the logic if it got lost; however there should be code elsewhere which uses this field and which are not being added in this PR.

pex.go
```diff
@@ -144,8 +145,11 @@ func (me *pexMsgFactory) PexMsg() *pp.PexMsg {
...
+	hold []pexEvent // delayed drops
```
yarikk, line 149: This one's also unrelated to the throttling: it's needed to maintain the minimum of 25 always visible peers when the current connection count is below 25, also as provisioned in the BEP. Also good to recover the logic if it got lost. There also should be code elsewhere which uses this field and which are not being added in this PR.

pex.go
```diff
@@ -144,8 +145,11 @@ func (me *pexMsgFactory) PexMsg() *pp.PexMsg {
...
+	// Torrent-wide cooldown deadline on inbound. This exists to prevent PEX from drowning out other
+	// peer address sources, until that is fixed.
+	rest time.Time
```
yarikk, line 152: Yep, that's the throttling timer.

yarikk (commented):
LGTM in the throttling part. There seem to be couple of other things being brought back here, albeit I'm not too sure those are indeed complete. See my inline comments.

Issue #845, opened 2023-07-03
udp tracker panic


monkeyWie:
When I add some tracker URLs, it causes a panic:

```
panic: read udp4 0.0.0.0:51174: wsarecvfrom: A message sent on a datagram socket was larger than the internal message buffer or some other network limit, or the buffer used to receive a datagram into was smaller than the datagram itself.

goroutine 1911 [running]:
github.com/anacrolix/torrent/tracker/udp.(*ConnClient).reader(0xc001b3ec80)
        D:/data/go/pkg/mod/github.com/anacrolix/torrent@v1.46.0/tracker/udp/conn-client.go:42 +0x4c5
created by github.com/anacrolix/torrent/tracker/udp.NewConnClient
        D:/data/go/pkg/mod/github.com/anacrolix/torrent@v1.46.0/tracker/udp/conn-client.go:102 +0x3e5

```

I found this issue https://github.com/anacrolix/torrent/issues/764, and upgraded to the latest version, but it still panics.

reproduce  code:

```go
package main

import (
	"github.com/anacrolix/torrent"
)

func main() {
	c, _ := torrent.NewClient(nil)
	defer c.Close()
	t, _ := c.AddMagnet("magnet:?xt=urn:btih:ca21e5baa68fe0edb715f5edf5120aa643218391&dn=[电影天堂www.dytt89.com]铃芽之旅-2022_HD日语中字.mp4")
	t.AddTrackers([][]string{{"udp://tracker.theoks.net:6969/announce"}})
	<-t.GotInfo()
}

```

anacrolix:
Thank you for reporting this.

PR #846, opened 2023-07-04
fix: udp tracker panic

monkeyWie:
fix https://github.com/anacrolix/torrent/issues/845

Some traker servers do not follow the udp traker specification, and the response packet returned is larger than the defined udp buffer size. We need to handle the error instead of panic.


anacrolix:
Thanks for the fix. I really don't like the error we get here, but I can't argue with the improved behaviour.

Issue #851, opened 2023-07-20
file.BytesCompleted() always return 0


monkeyWie:
Firstly thanks for your great project. There have a problem that when I call `file.BytesCompleted()`  to stat a special file speed it always return 0, bur call `torrent.BytesCompleted()` is fine, and it only seems to happen in torrent that piece length is longer.

This is my code:
```
fmt.Printf("file[%d] %d %d\n", i, torrent.Files()[i].BytesCompleted(), torrent.BytesCompleted())
// output
file[0]: 0 16384
file[0]: 0 65536
file[0]: 0 114688
file[0]: 0 245760
```

anacrolix:
Sorry, it took a while for me to understand the problem. File.BytesCompleted is not including dirty chunks in the computation, but it does for Torrent.BytesCompleted.

anacrolix:
One reason not to include dirty bytes in the computation is that they can fail piece checks, and so your download rate would appear to go backwards between successive calls. Is that okay?

PR #852, opened 2023-07-22
fix: torrent file real time completed bytes

monkeyWie:
fixed https://github.com/anacrolix/torrent/issues/851.

### DIFF
--- a/file.go
+++ b/file.go
@@ -62,31 +62,58 @@ func fileBytesLeft(
 	fileTorrentOffset int64,
 	fileLength int64,
 	torrentCompletedPieces *roaring.Bitmap,
+	pieceSizeCompletedFn func(pieceIndex int) int64,
 ) (left int64) {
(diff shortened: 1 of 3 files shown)


anacrolix:
Thanks for the contribution. This really needs some tests. I'm also not sure about allocating a roaring bitmap to perform this. Are we sure there's not just a simple error in the existing function?

anacrolix:
I've had time to think about this carefully. This is a pretty good solution. On the boundary pieces, it is actually possible to figure out if dirty chunks belong to the file or not. I see that you don't do this, but you do apply a limit to ensure that the dirty chunk attribution doesn't exceed the amount of the file that is inside the boundary piece. It would take me quite a while to patch it up to handle the chunk attribution to be 100% correct, so I will merge this as you have it, and accept or add the last mile if it comes up again.

I do note the existing tests don't do any special dirty chunk treatment, it would be nice to augment the tests a bit to handle this too.

PR #860, opened 2023-08-23
don't return error when trying to read past EOF

frrad:
I encountered a bug where when an application attempts to read "across" the EOF (i.e. try to read $x$ bytes starting at offset $y$ where $x+y > len(file)$ ). The application dies with "unexpected EOF".

This patch fixes the bug for me.

I suspect that the fuse library does not expect `Read` to return non-nil errors except when no data was able to be read, but wasn't able to get into that library enough to confirm.


anacrolix:
Do you have more context for your change @frrad ?

frrad:
whoops! updated

anacrolix:
Bummer. That violates Go's contract here, but should really be fixed in the fuse wrapper. The tests are passing so let's go with it. The fuse wrapper is at https://github.com/anacrolix/fuse if you wanted to try to find it.

PR #870, opened 2023-09-16
simply defer close

ucwong:


anacrolix:
I don't think this is straightforward. On Windows, I believe you need to keep the file open while you have memory maps into it. Take a look at this for example: https://github.com/anacrolix/torrent/commit/dee985f9bdd318e52edb90ebeb0cd787a0505a78.

PR #871, opened 2023-09-16
avoid using buf reset

ucwong:


metainfo/metainfo.go
```diff
@@ -44,9 +44,8 @@ func LoadFromFile(filename string) (*MetaInfo, error) {
 		return nil, err
 	}
 	defer f.Close()
-	var buf bufio.Reader
-	buf.Reset(f)
-	return Load(&buf)
+	buf := bufio.NewReader(f)
+	return Load(buf)
```
anacrolix, line 48: I did this to try to avoid an extra allocation for the Reader.

anacrolix:
I see you're keen to contribute @ucwong, is there something specific you are looking to do?

ucwong:
> I see you're keen to contribute @ucwong, is there something specific you are looking to do?

PR will be created once I find issues in this repo

PR #884, opened 2023-12-04
bencode: drop support for go 1.20

vassilit:


anacrolix:
I appreciate the tidy up, but I prefer to make it easier to backport if needed, rather than tighten the constraints, at least for something like this that has no runtime or compile time overhead.

vassilit:
I wouldn't have tidied up either it if it wasn't for commit e10ffd133 « Drop support for Go 1.20 ».
I thought you forgot to clean up here as well.

anacrolix:
Yeah it was more that it's a hassle to guarantee backward compatibility, but I don't want to throw it away.

Your contributions are welcome, thanks for the consideration.

Issue #369, opened 2020-01-13
Clean up torrent creation methods on Client


anacrolix:
Per https://github.com/anacrolix/torrent/issues/243#issue-310328823:

> remove torrent creation methods from the client. there are just want too many of them. poorly documented, and its unclear which one should be used and when.
> ```
> func (cl *Client) AddMagnet(uri string) (T *Torrent, err error) {}
> func (cl *Client) AddTorrentFromFile(filename string) (T *Torrent, err error) {}
> func (cl *Client) AddTorrent(mi *metainfo.MetaInfo) (T *Torrent, err error) {}
> func (cl *Client) AddTorrentInfoHash(infoHash metainfo.Hash) (t *Torrent, new bool) {}
> func (cl *Client) AddTorrentInfoHashWithStorage(infoHash metainfo.Hash, specStorage storage.ClientImpl) (t *Torrent, new bool) {}
> func (cl *Client) AddTorrentSpec(spec *TorrentSpec) (t *Torrent, new bool, err error) {}
> ```

Issue #243, opened 2018-04-01
API cleanup suggestions


james-lawrence:
as discussed in #218, the API is a little hard to follow and use for certain use cases. The goal here is to discuss and suggest usability improvements to the API. I believe these changes could also be used to help resolve #238, #168 in a clean fashion.

- remove torrent creation methods from the client. there are just want too many of them. poorly documented, and its unclear which one should be used and when.

```golang
func (cl *Client) AddMagnet(uri string) (T *Torrent, err error) {}
func (cl *Client) AddTorrentFromFile(filename string) (T *Torrent, err error) {}
func (cl *Client) AddTorrent(mi *metainfo.MetaInfo) (T *Torrent, err error) {}
func (cl *Client) AddTorrentInfoHash(infoHash metainfo.Hash) (t *Torrent, new bool) {}
func (cl *Client) AddTorrentInfoHashWithStorage(infoHash metainfo.Hash, specStorage storage.ClientImpl) (t *Torrent, new bool) {}
func (cl *Client) AddTorrentSpec(spec *TorrentSpec) (t *Torrent, new bool, err error) {}
```

- Create new standalone methods for the creation of torrent specifications. these methods have no concept of the client and just know how to create torrent specs.

```golang
type TorrentSpecOption func(*TorrentSpec)
// TorrentSpecData provide the source data to the torrent, nil can represent no local data.
type TorrentSpecData(d TorrentImpl) TorrentSpecOption {
  return func(t *TorrentSpec) {
    t.TorrentImpl = d
  }
}

// New create a torrent from the metainfo.Info and any additional options.
func New(mi metainfo.Info, options ...TorrentSpecOption) (*TorrentSpec, error) {}

// NewFromData creates a torrent from the raw data the torrent will represent.
func NewFromData(r io.Reader, options ...TorrentSpecOption) (*TorrentSpec, error) {}

// NewFromFile convience method to create a torrent from a file. see NewFromData for
// more information.
func NewFromFile(path string, options ...TorrentSpecOption) (*TorrentSpec, err error) {}

// NewFromMetadata creates a torrent from metadata stored in a file based format.
func NewFromMetadata(r io.ReadCloser, options ...TorrentSpecOption) (*TorrentSpec, error) {}

// NewFromMagnet create a torrent from a magnet URL.
func NewFromMagnet(s string, options ...TorrentSpecOption) (*TorrentSpec, error) {}
```

-  Add two new methods to the client that just deal with torrent specs.
```golang
[...]

PR #887, opened 2024-01-07
Refinement of API Design with Introduction of AddTorrent2 Function

kazhuravlev:
Related with #369 

I have carefully reviewed the discussions in thread #243 regarding the design of a new API, but it appears that a definitive decision has not been reached. In response, I have proposed a straightforward solution that maintains simplicity and allows for future refactoring without introducing breaking changes.

**Key Changes:**

- Introducing a new function named `AddTorrent2` to address the absence of an available name for the existing `AddTorrent` function. (do you have any idea about a new name?)
- All pre-existing functions now utilize the newly introduced `AddTorrent2` function.
- Deprecating all previous functions to guide users toward adopting the updated design.
- Incorporating the use of `context.Context` within the `AddTorrent2` function to facilitate potential internal library use in the future.
- Implementing the `addTorrentReq` interface and its associated implementations to provide flexibility for future design modifications without causing disruptions for end-users.

**Important Note:**

It has been observed that `AddTorrentInfoHashWithStorage` bears a striking resemblance to `AddTorrentOpt`, with the only discernible difference being the absence of `t.setInfoBytesLocked(opts.InfoBytes)` in the new version. While I believe this modification is secure, it is crucial to bring attention to this point for thorough consideration.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1832,3 +1832,80 @@ func (cl *Client) addTorrentFromSpec(spec *TorrentSpec) (*Torrent, bool, error) 
 
 	return t, isNew, nil
 }
+
+// addTorrentReq hide all internal details from consumers of AddTorrent2 and allows you
+// to change it to any other style/way without breaking changes.
+type addTorrentReq interface {
+	isAddTorrentReq()
+}
+
+type rMagnetURI struct {
+	URI string
+}
+
+func (rMagnetURI) isAddTorrentReq() {}
+
+func FromMagnetURI(uri string) rMagnetURI {
+	return rMagnetURI{
+		URI: uri,
+	}
+}
+
+type rTorrentSpec struct {
+	Spec *TorrentSpec
+}
+
+func (rTorrentSpec) isAddTorrentReq() {}
+
+func FromTorrentSpec(spec *TorrentSpec) rTorrentSpec {
+	return rTorrentSpec{
+		Spec: spec,
+	}
+}
+
+type rTorrentOpts struct {
+	Opts AddTorrentOpts
+}
+
+func (rTorrentOpts) isAddTorrentReq() {}
+
+func FromTorrentOpts(opts AddTorrentOpts) rTorrentOpts {
+	return rTorrentOpts{
+		Opts: opts,
+	}
+}
+
+type rHash struct {
+	Hash metainfo.Hash
+}
+
+func (rHash) isAddTorrentReq() {}
+
+func FromHash(hash metainfo.Hash) rHash {
+	return rHash{Hash: hash}
+}
+
+type rMetaInfo struct {
+	Meta *metainfo.MetaInfo
+}
+
+func (rMetaInfo) isAddTorrentReq() {}
+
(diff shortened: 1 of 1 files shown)


anacrolix:
Thanks for looking into this. The issue mentions that AddTorrentOpt was the intended way forward.

In particular there are 2 parts to the API: Things that must be set when a Torrent is added, and things that can be altered after adding. AddTorrentOpt was the first part of this, and the rest were to be made their own methods on Torrent.

anacrolix:
If you could reduce the diff by moving the methods to their original locations that would make reviewing the changes a lot easier.

kazhuravlev:
So what's about AddTorrentOpt-should i change something right now? I am asking because in case with TorrentOpt struct you will have a god object that contains conflicted fields. For example InfoHash,Filename,MetaInfo and others. with this big struct we should check that only one of "source" fields are filled. Do i understand right?

kazhuravlev:
@anacrolix I suppose the new diff looks better. But I am not sure what you decide about the mechanism of torrent creation. If you expect any changes in PR - just let me know.

anacrolix:
I see what you're going for now. If by god object you mean "configuration" style structs, unfortunately they're a reality in Go that are easier when you embrace them. I'd refer you to examples like TLS.Config. They are grating, but they do actually work surprisingly well in the context of Go. I think that is where I was going with AddTorrentOpt (and unintentionally with TorrentSpec, which I just don't like it's too big and crosses between initialization and stuff that can be configured later. Your PR makes me think about the issue of having magnet links, torrent files, and infohashes and how having fields for each of those would make for a horrible config struct. To that end I think the API has a few parts: being able to unpack special representations (we have this for MetaInfo and Magnet). Adding torrents and magnet links to a Client (we have helpers for this, plus TorrentSpec which can be built from the earlier representations). Then the separation of things that must occur at initialization and things that can change at runtime. The last point is the interesting area: things that occur at initialization we have AddTorrentOpt, and I intentionally want to avoid having a config struct for runtime things. Go isn't good for that again: we can't easily snapshot the current runtime config, make changes and push it back in without all kinds of silliness. Maybe we could, but also people don't particularly like that style. How this maps back to the "do-all" intermediate configuration (currently TorrentSpec) is ugly. If you make changes to TorrentSpec, some of them can't modified at runtime, so it's kind of broken.

kazhuravlev:
Yes, I see you understand me. Right now you have all the options to implement AddTorrent (or similar function). But I have to say that it’s not entirely clear to me – did you make any decisions about the design or was it just thinking out loud?

In any case, feel free to write some specific goals that I can achieve. Also, if it’s faster than describing the tasks - you can merge this PR and rewrite it yourself. 

I just need a safe and clear API as final user.

anacrolix:
Sorry there's no progress on this. There's a very long list of downstream users, and it's not a simple case of tearing down a lot of the deprecated API. The API has been very stable for 7 years or so, so I wouldn't worry about it changing on you.

Issue #897, opened 2024-02-02
Slow downloads while multiple large torrents are being downloaded.


veshij:
We are observing extremely slow download speeds when multiple large torrents are being downloaded.
We embed library in our application with following configuration:
```
	config := torrent.NewDefaultClientConfig()
	config.DhtStartingNodes = func(string) dht.StartingNodesGetter { return func() ([]dht.Addr, error) { return nil, nil } }
	config.UploadRateLimiter = rate.NewLimiter(32*1024*1024, 64*1024*1024)
	config.DownloadRateLimiter = rate.NewLimiter(128*1024*1024, 256*1024*1024)
	config.DialRateLimiter = rate.NewLimiter(128, 256)
	config.EstablishedConnsPerTorrent = 128
	config.DisableUTP = true
	config.NoDHT = true
	config.Seed = true
	config.DisableWebseeds = true
	config.ListenPort = 5502
```
All torrents have a piece size 256k.
Some torrents distribute 5-6GB datafiles, some are fairly small, hundreds of MB.

We observe that when "large" torrent is being downloaded - this torrent and other torrents processed by the same process are downloaded way below DL/UL rate-limit speeds. Once large torrent is completed - "smaller" torrents are processed with regular/expected speeds.

Looks like most of the CPU time is spent in request-strategy.GetRequestablePieces:

<img width="1707" alt="Screenshot 2024-02-02 at 10 29 51 AM" src="https://github.com/anacrolix/torrent/assets/6113489/707d60d2-4765-4fd1-a2e3-be7b4c5b7af6">

[pprof.gz](https://github.com/anacrolix/torrent/files/14143930/07516473769856156647__CPU__neptune.packagestore__iad__611ea46cd5e212f1fb40dc3003d9711377c15353-961fef5d87584408b4b4b83e53ce72bd__iad9a-rf13-12b.pprof.gz)

Could  you suggest what we can change to resolve the issue?
Thank you!

anacrolix:
Thanks for the report. What version of the client are you using?

Some workarounds include: Limiting the number of pieces per torrent. There is an algorithm that does this include in this implementation. Another is to add a capacity to your storage. I think this will short circuit how deep the implementation consists potential requests. See the storage package interfaces for this.

Feel free to share a list of torrents that make the issue reproducible.

veshij:
We use `v1.53.2`.
>Limiting the number of pieces per torrent.
[...]

PR #901, opened 2024-02-08
Stop PRO tree traversal when allTorrentsUnverifiedBytes reaches MaxUnverifiedBytes

veshij:
If I understand correctly we should stop processing tree items once MaxUnverifiedBytes limit is reached.
Existing implementation still goes through the whole tree resolving pointer to the torrent and calling t.IgnorePiece(_i.key.Index) for each piece.

while here moving storage check at the very end since it's quite unlikely it'll return False compared to MaxUnverifiedBytes or t.IgnorePiece().

### DIFF
--- a/request-strategy/order.go
+++ b/request-strategy/order.go
@@ -67,22 +67,22 @@ func GetRequestablePieces(
 
 		ih := _i.key.InfoHash
 		t := input.Torrent(ih)
+		pieceLength := t.PieceLength()
+		if storageLeft != nil {
+			if *storageLeft < pieceLength {
+				return false
+			}
+			*storageLeft -= pieceLength
+		}
 		if t.IgnorePiece(_i.key.Index) {
 			// TODO: Clarify exactly what is verified. Stuff that's being hashed should be
 			// considered unverified and hold up further requests.
 			return true
 		}
 
-		pieceLength := t.PieceLength()
 		allTorrentsUnverifiedBytes += pieceLength
 		if input.MaxUnverifiedBytes() != 0 && allTorrentsUnverifiedBytes > input.MaxUnverifiedBytes() {
 			return false
-		}
-		if storageLeft != nil {
-			if *storageLeft < pieceLength {
-				return false
-			}
-			*storageLeft -= pieceLength
 		}
 
 		f(ih, _i.key.Index, _i.state)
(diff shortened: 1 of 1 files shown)


anacrolix:
I mentioned the short circuiting in https://github.com/anacrolix/torrent/issues/897#issuecomment-1931141820.

I believe it's correct to return false for that case. I don't think it will be faster (or correct) to move the other stuff around.

request-strategy/order.go
```diff
@@ -67,22 +67,24 @@ func GetRequestablePieces(
 
 		ih := _i.key.InfoHash
 		t := input.Torrent(ih)
+		if t.IgnorePiece(_i.key.Index) {
```
anacrolix, line 70: I don't believe this can be moved up like this.

request-strategy/order.go
```diff
@@ -67,22 +67,24 @@ func GetRequestablePieces(
...
+			// TODO: Clarify exactly what is verified. Stuff that's being hashed should be
+			// considered unverified and hold up further requests.
+			return true
+		}
+
 		pieceLength := t.PieceLength()
+		allTorrentsUnverifiedBytes += pieceLength
+		if input.MaxUnverifiedBytes() != 0 && allTorrentsUnverifiedBytes > input.MaxUnverifiedBytes() {
+			return false
```
anacrolix, line 79: As mentioned elsewhere, I think this is correct.

request-strategy/order.go
veshij: hm, could you help me understand why not?
Previous code:
- get torrent
- check if storage is full (quite unlikely)
- chef if piece should be ignored (quite likely)

Updated:
- get torrent
- chef if piece should be ignored (quite likely)
- check if storage is full (quite unlikely)
I don't think there is drastic difference, but feels like It can save a couple of cpu cycles in cases with millions of pieces, I don't mind discarding this change.

anacrolix: Yes, let me try:

If a piece is ignored (not wanted, which doesn't preclude already having it), it may still have a priority that makes it desirable to occupy storage space. So we want to the piece against the available storage capacity, even if we don't need to request it (because we already have it most likely).

Additionally, the change you made move the unverified bytes check before the total storage check. I don't think this has any impact on behaviour (assuming it returns false like the storage capacity check). I wonder which is faster.

Thanks for your efforts to look into this.

PR #902, opened 2024-02-09
GetRequestablePieces: remove runtime correctness check

veshij:
In profiles I collected correctness check https://github.com/anacrolix/torrent/blob/c73c99b372114204b3e62ab7fe7103b2477c7dee/request-strategy/order.go#L60-L66 takes up to 10% cpu time. Would it make sense to remove it from runtime and add a unit-test for that instead (if not already)?

Screenshot 2024-02-07 at 11 25 12 PM

additional discussion is in https://github.com/anacrolix/torrent/issues/897#issuecomment-1933596287

### DIFF
--- a/request-strategy/order.go
+++ b/request-strategy/order.go
@@ -4,7 +4,6 @@ import (
 	"bytes"
 	"expvar"
 
-	g "github.com/anacrolix/generics"
 	"github.com/anacrolix/multiless"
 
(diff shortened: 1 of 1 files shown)


anacrolix:
I tried this on top of https://github.com/anacrolix/torrent/issues/897#issuecomment-1935323620, and it knocked off another 10%. This can be merged independently of evaluating that other change. Thanks for the contribution!

PR #920, opened 2024-03-25
Support ICE servers auth

### LINKED ISSUE #903: ICE Servers config doesn't support authentication for TURN servers
In #824 I added the option to configure the client with a list of ICE servers. The list is of type `[]string` and is used to provide the URLs of the servers to the WebRTC peer connections, [here](https://github.com/anacrolix/torrent/blob/master/webtorrent/transport.go#L56).

This will work fine when relying on STUN servers, which is the most common solution and doesn't usually require authentication.

However, on some occasions TURN servers are needed for NAT traversal. TURN servers normally require authentication, and the current implementation doesn't allow for that in the client config.

@anacrolix if interested, I'd be happy to open a PR to support passing ICE servers of type `[]webrtc.ICEServer` as used in Pion. I'm just unsure whether there's a clean way of doing it without introducing a breaking change?

marcovidonis:
Support passing ICE servers of type `[]webrtc.ICEServer` to the client config. This effectively enables the use of TURN servers when NAT traversal is needed. TURN servers normally require authentication, and the current implementation doesn't allow for that in the client config.

Fixes #903.

### DIFF
--- a/config.go
+++ b/config.go
@@ -185,9 +185,13 @@ type ClientConfig struct {
 
 	Callbacks Callbacks
 
-	// ICEServers defines a slice describing servers available to be used by
+	// ICEServerList defines a slice describing servers available to be used by
 	// ICE, such as STUN and TURN servers.
-	ICEServers []webrtc.ICEServer
+	ICEServerList []webrtc.ICEServer
+
+	// Legacy support. ICEServers does not support server authentication and therefore
+	// it cannot be used with most TURN servers. Use ICEServerList instead.
+	ICEServers []string
 
 	DialRateLimiter *rate.Limiter
 
--- a/config.go
+++ b/config.go
@@ -7,6 +7,8 @@ import (
 	"net/url"
 	"time"
 
+	"github.com/pion/webrtc/v3"
+
 	"github.com/anacrolix/dht/v2"
 	"github.com/anacrolix/dht/v2/krpc"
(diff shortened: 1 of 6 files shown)


config.go
```diff
@@ -183,8 +185,12 @@ type ClientConfig struct {
 
 	Callbacks Callbacks
 
-	// ICEServers defines a slice describing servers available to be used by
+	// ICEServerList defines a slice describing servers available to be used by
 	// ICE, such as STUN and TURN servers.
+	ICEServerList []webrtc.ICEServer
+
+	// Legacy support. ICEServers does not support server authentication and therefore
```
anacrolix, line 192: Could you add "Deprecated"? It flags in most doc and editors

webtorrent/tracker-client.go
```diff
@@ -9,6 +9,8 @@ import (
 	"sync"
 	"time"
 
+	"github.com/anacrolix/torrent/types/infohash"
```
anacrolix, line 12: I do believe this isn't needed, it's causing the CI failures

anacrolix (changes requested):
This is really good, thanks. Let me know if you don't have time to make the changes and I'll finish it. If you do have time, run https://github.com/anacrolix/gorond over the repo, it will fix up your imports.

anacrolix:
Thank you sir

PR #922, opened 2024-04-01
Ignore pieces with availability of 0

aritra24:
resolves: #916 
Pieces with zero availability shouldn't
consume MaxUnverifiedBytes and shouldn't
be prioritised.

### DIFF
--- a/request-strategy/order.go
+++ b/request-strategy/order.go
@@ -67,6 +67,9 @@ func GetRequestablePieces(
 			// considered unverified and hold up further requests.
 			return true
 		}
+		if _i.state.Availability == 0 {
+			return false
+		}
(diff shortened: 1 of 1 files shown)


anacrolix:
Thanks for finding this. The solution isn't quite correct. I am mucking around with a more comprehensive solution at https://github.com/anacrolix/torrent/tree/herp.

Specifically I think it's necessary to ensure requests to individual peers aren't starved by pieces that aren't available to from them.

aritra24:
Ah that's fair, I wasn't too confident on the solution myself. I can close this pr then. Thanks for helping figure it out ✌🏼

anacrolix:
Thank you for persisting and finding the issue! I'll have a fix in master shortly.

anacrolix:
Fixed in 528c5f69d657d6edb7359674473eec1ccff10b9a

PR #945, opened 2024-05-23
add ClearTrackers() func

ucwong:


anacrolix:
I think this is missing the removal of the active tracker announcer instances. I also think it's probably that there should just be a Torrent.ModifyTrackers that receives the current tracker list, you can alter it, then it updates the announce instances in one atomic change.

PR #947, opened 2024-05-26
clone func used since go 1.21

ucwong:


anacrolix:
I don't think this is the same behaviour as the original. You need to be cloning the internal slices.

ucwong:
> I don't think this is the same behaviour as the original. You need to be cloning the internal slices.

```
import (
        "fmt"
        "golang.org/x/tools/go/packages"
)

func main() {
        original := [][]string{
                {"a", "b"},
                {"c", "d"},
        }

        // Deep copy using slices.Clone
        cloned := slices.Clone(original)

        // Modify the original
        original[0][0] = "X"

        fmt.Println("Original:", original) // Output: [[X, "b"], ["c", "d"]]
        fmt.Println("Cloned:", cloned)  // Output: [['a', 'b'], ['c', 'd']]
}
```
In this example:

original is a slice of slices containing two inner slices of strings.
cloned is created using slices.Clone(original), resulting in a deep copy.
Modifying original doesn't affect cloned because they are independent copies.
Key Points:

slices.Clone is available in Go 1.21 and later.
For earlier versions, you can use the golang.org/x/exp/slices package (experimental).
This approach is efficient and avoids manual copying loops.
I hope this explanation clarifies how slices.Clone works with [][]string in Go!

PR #957, opened 2024-06-30
remove golang.org/x/exp since it doesn't follow go 1 compatibility promise

trim21:
[document](https://pkg.go.dev/golang.org/x/exp#section-readme) says `golang.org/x/exp` doesn't follow go 1 compatibility promise, so copy the package we used to internal for future compatibility promise.

In the future if downsteam user import exp with a newer non-compatibility version it will get unexpected surprise.

it's still a indirect dependency from `github.com/anacrolix/generics`, so I suggest to remove it from `github.com/anacrolix/generics` too

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -52,7 +52,6 @@ require (
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.11.1
 	go.opentelemetry.io/otel/sdk v1.11.1
 	go.opentelemetry.io/otel/trace v1.11.1
-	golang.org/x/exp v0.0.0-20240506185415-9bf2ced13842
 	golang.org/x/sync v0.7.0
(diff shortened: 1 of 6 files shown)


anacrolix:
I'm really not that worried about it. If exp removes the APIs, it still won't break people unless they're aggressively updating their transitive dependencies (a common anti-pattern in Go, `go get -u` specifically).

That said, the Go team has been known to remove APIs for pedantic reasons, but it's still not that worrying.

trim21:
downstream users may rely on another package non-compatibility exp versions

trim21:
go1.23 will add these 2 functions to stdlib

anacrolix:
Awesome, feel free to make a PR to use the functions from the standard library

anacrolix:
I fixed this in master, thanks!

PR #960, opened 2024-07-10
use bufio for mse write

trim21:
there is no need to start a new goroutine for writing small chunks.

If you want concurrently read write we can use chan, using slice, lock and cond to mock a chan doesn't sound like a good idea.

### DIFF
--- a/mse/mse.go
+++ b/mse/mse.go
@@ -3,6 +3,7 @@
 package mse
 
 import (
+	"bufio"
 	"bytes"
 	"crypto/rand"
 	"crypto/rc4"
@@ -15,9 +16,6 @@ import (
 	"math"
 	"math/big"
 	"strconv"
-	"sync"
-
-	"github.com/anacrolix/missinggo/perf"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


anacrolix:
It's very old code, I'm sure I'd do it differently now. 

Channels are usually over used in Go, they're also much worse for performance than people realise. I doubt this situation is affected. 

Could you provide a benchmark to demonstrate the improvement? 

Thanks for the contribution!

trim21:
before:
cpu: Intel(R) Pentium(R) Gold G6405 CPU @ 4.10GHz
BenchmarkMSE-4              4141            301669 ns/op           35309 B/op        220 allocs/op

after:
cpu: Intel(R) Pentium(R) Gold G6405 CPU @ 4.10GHz
BenchmarkMSE-4              4305            292442 ns/op           42028 B/op        207 allocs/op

trim21:
emm I guess current implment is right. using bufio may cause a dead lock problem.

We should create a new goroutine to do the writing.

anacrolix:
Thanks for the effort and the conclusion. It's important that people test assumptions like this.

PR #968, opened 2024-08-26
Basic support for serializing v2 torrent file

Mivik:
Implements bencode.Marshaler for FileTree

Omit Pieces field in torrent info if empty


anacrolix:
I completely forgot about this. Thanks!

PR #969, opened 2024-08-29
Basic support for responding to HashRequest

Mivik:
This PR implemented serialization of Hashes and HashReject messages, and the logic of responding to HashRequest. Note that some code relevant to `Piece` is changed and should be reviewed.

### DIFF
--- a/peer_protocol/msg.go
+++ b/peer_protocol/msg.go
@@ -67,18 +67,19 @@ type MessageWriter interface {
 	io.Writer
 }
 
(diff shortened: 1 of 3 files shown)


anacrolix:
Thanks! Nice implementation. The implementation of hasPieceLayer looks correct. You've constrained the checks to v2 fields in the info but that makes sense. Were you seeing performance issues with the old implementation, or just tidying up?

anacrolix:
Ok I tested it pretty well.

PR #983, opened 2024-09-30
Expose WebRTC peerconn stats

marcovidonis:
I revisited the work on #803 and improved on it. By saving a pointer to the open peer connections, we get stats more reliably. Stats now also work when seeding.

A note on the `Pion/webrtc` version. While some stats are available in the current latest version (`v3`), the project greatly improved on them in `v4`, currently in beta. I'd recommend upgrading to `v4` as soon as it's convenient when it gets out of beta.

### DIFF
--- a/torrent.go
+++ b/torrent.go
@@ -3003,16 +3003,15 @@ func (t *Torrent) iterUndirtiedRequestIndexesInPiece(
 	)
 }
 
-type webRtcStatsReports []webrtc.StatsReport
+type webRtcStatsReports map[string]webrtc.StatsReport
 
 func (t *Torrent) GetWebRtcPeerConnStats() map[string]webRtcStatsReports {
 	stats := make(map[string]webRtcStatsReports)
 	trackersMap := t.cl.websocketTrackers.clients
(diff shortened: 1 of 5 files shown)


anacrolix (approved):

marcovidonis:
That's great, thanks! Yeah I don't really see how to make unit tests for this... we'd likely need an integration test of some kind.

I think the PR is almost ready now, as it does expose the stats we get from Pion. I noticed the wasm test fails because Pion doesn't yet support stats on wasm. Do you know of a way I could avoid calling `GetStats()` if we're on a wasm architecture?

anacrolix:
I'd probably put the GetStats behind a build tag that matches whatever pion uses. For the unsupported tags, return something appropriate.

marcovidonis:
Thanks for that, @anacrolix! All checks pass now, so I think you can merge if you're happy with this.

Also, I think a long-lived branch for Pion v4 could be useful, so I'd be happy to make one!

anacrolix:
Go for it, thanks!

Issue #982, opened 2024-09-30
Speed too slowly if set cpu-limit for pod in Kubernetes. Or high cpu usage with cpu-nolimit.


penglongli:
I want to develop a file distributor based on torrent and deploy it in a Kubernetes cluster using DaemonSet. I deployed the [chihaya/chihaya](https://github.com/chihaya/chihaya) as the tracker.

Every node i have a torrent client:
```
	clientConfig := torrent.NewDefaultClientConfig()
	clientConfig.DataDir = th.op.TransferPath
	clientConfig.Seed = true
	clientConfig.ListenPort = int(th.op.TorrentPort)
	clientConfig.DefaultStorage = storage.NewFileByInfoHash(th.op.TransferPath)
	clientConfig.DisableUTP = true
	clientConfig.MaxUnverifiedBytes = 64 << 30
	clientConfig.NoDHT = false
	clientConfig.DisablePEX = false
	clientConfig.EstablishedConnsPerTorrent = 200
	clientConfig.HalfOpenConnsPerTorrent = 100
	clientConfig.TorrentPeersHighWater = 2000
	clientConfig.MaxUnverifiedBytes = 671088640
```

When i need distribute a file from one node, i'll generate the torrent first
```
pieceLength = metainfo.ChoosePieceLength(layerFileInfo.Size())
info := metainfo.Info{
	PieceLength: pieceLength,
}
if err := info.BuildFromFilePath(layerFile); err != nil {
	return nil, errors.Wrapf(err, "build torrent metainfo from file '%s' failed", layerFileInfo)
}
mi := metainfo.MetaInfo{
	InfoBytes: bencode.MustMarshal(info),
}
ih := mi.HashInfoBytes()
to, _ := th.client.AddTorrentOpt(torrent.AddTorrentOpts{
	InfoHash: ih,
	Storage: storage.NewFileOpts(storage.NewFileClientOpts{
		ClientBaseDir: layerFile,
		FilePathMaker: func(opts storage.FilePathMakerOpts) string {
			return filepath.Join(opts.File.BestPath()...)
		},
		TorrentDirMaker: nil,
		PieceCompletion: th.pc,
	}),
	ChunkSize: 128 * 1024,
})
err := to.MergeSpec(&torrent.TorrentSpec{
	InfoBytes: mi.InfoBytes,
	Trackers:  [][]string{{th.op.TorrentAnnounce}},
})
```

[...]

PR #985, opened 2024-10-14
Upgrade pion/webrtc to v4

marcovidonis:
`pion/webrtc/v4` got out of beta and was officially released last week. Looking at the breaking changes in the [release notes](https://github.com/pion/webrtc/wiki/Release-WebRTC@v4.0.0), I don't think there should be impacts on this project. All tests pass when I run them locally.

In #983 we decided to keep a separate branch for v4. I thought I'd also open a PR for it since it's no longer in beta. Are there any other checks that we could do so when know when we're ready to merge?


anacrolix:
No, if you're happy with it and it passed your checks that should be good!

anacrolix:
There is a performance issue that might be worth revisiting that exists because of a WebRTC workaround: https://github.com/anacrolix/torrent/issues/982#issuecomment-2402036313. I dont't know if you think this might fix it, or want to pion/v4 without that workaround and see if we can remove it?

marcovidonis:
I don't know if v4 will fix it, but it's worth a try! I'll give it a look

PR #989, opened 2024-11-18
Implement upload limit burst to 16 KiB

marco-ragusa:


marco-ragusa:
I edited it to use 16 KB, but maybe 64 KB is better.

anacrolix:
Was there some situation this improved for you?

PR #990, opened 2024-12-06
CI: go 1.22

AskAlexSharov:


anacrolix:
I think we're already on Go 1.22 on CI. I think this PR was by accident?

PR #998, opened 2025-02-28
adds bep_0014 (local peer discovery)

philrhc:
upstreams BEP14 from https://gitlab.com/axet/libtorrent/-/blob/master/lpd.go

### DIFF
--- a/bep14.go
+++ b/bep14.go
@@ -1,0 +1,369 @@
+package torrent
+
+import (
+	"bufio"
+	"bytes"
+	"fmt"
+	"log"
+	"net"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/anacrolix/missinggo"
+	"github.com/anacrolix/sync"
+	"github.com/anacrolix/torrent/metainfo"
+)
+
+var lpd *LPDServer
+var mu sync.Mutex
+
+// http://bittorrent.org/beps/bep_0014.html
+// TODO http://bittorrent.org/beps/bep_0026.html
+
+const (
+	bep14_host4    = "239.192.152.143:6771"
+	bep14_host6    = "[ff15::efc0:988f]:6771"
+	bep14_announce = "BT-SEARCH * HTTP/1.1\r\n" +
+		"Host: %s\r\n" +
+		"Port: %s\r\n" +
+		"%s" +
+		"\r\n" +
+		"\r\n"
+	bep14_announce_infohash = "Infohash: %s\r\n"
+	bep14_long_timeout      = 1 * time.Minute
+	bep14_short_timeout     = 2 * time.Second // bep14 - 1 minute. not practial. what if use start/stop another torrent? so make it 2 secs.
+	bep14_max               = 0               // maximum hashes per request, 0 - only limited by udp packet size
+)
+
+type LPDConn struct {
+	stop  missinggo.Event
+	force missinggo.Event
+
+	network     string // "udp4" or "udp6"
+	addr        *net.UDPAddr
+	mcListener  *net.UDPConn
+	mcPublisher *net.UDPConn
+	host        string // bep14_host4 or bep14_host6
+}
+
+func lpdConnNew(network string, host string) *LPDConn {
+	m := &LPDConn{}
+
+	m.network = network
+	m.host = host
+
+	var err error
+
+	m.addr, err = net.ResolveUDPAddr(m.network, m.host)
+	if err != nil {
+		log.Println("LPD unable to start", err)
+		return nil
+	}
+	m.mcListener, err = net.ListenMulticastUDP(m.network, nil, m.addr)
+	if err != nil {
+		log.Println("LPD unable to start", err)
+		return nil
+	}
+	m.mcPublisher, err = net.DialUDP(network, nil, m.addr)
+	if err != nil {
+		fmt.Println("Error dialing UDP:", err)
+		return nil
+	}
+
+	return m
+}
+
+func (m *LPDConn) receiver(client *Client) {
+	for {
+		mu.Lock()
+		conn := m.mcListener
+		if conn == nil {
+			mu.Unlock()
+			return
+		}
+		mu.Unlock()
+
+		buf := make([]byte, 2000)
+		_, from, err := conn.ReadFromUDP(buf)
+		if err != nil {
+			log.Println("receiver", err)
+			continue
+		}
+
+		req, err := http.ReadRequest(bufio.NewReader(bytes.NewReader(buf)))
+		if err != nil {
+			log.Println("receiver", err)
+			continue
+		}
+
+		if req.Method != "BT-SEARCH" {
+			log.Println("receiver", "Wrong request: ", req.Method)
+			continue
+		}
+
+		// bep14 says here can be multiple response headers
+		var ihs []string = req.Header[http.CanonicalHeaderKey("Infohash")]
+		if ihs == nil {
+			log.Println("receiver", "No Infohash")
+			continue
+		}
+
+		port := req.Header.Get("Port")
+		if port == "" {
+			log.Println("receiver", "No port")
+			continue
+		}
+
+		addr, err := net.ResolveUDPAddr(m.network, net.JoinHostPort(from.IP.String(), port))
+		if err != nil {
+			log.Println("receiver", err)
+			continue
+		}
+
+		mu.Lock()
+		if lpd == nil { // can be closed already
+			mu.Unlock()
+			return
+		}
+		lpd.peer(addr.String())
+		lpd.refresh()
+		//log.Println("LPD", m.network, addr.String(), ih)
+		ignore := make(map[*Torrent]bool)
+		for _, ih := range ihs {
+			hash := metainfo.NewHashFromHex(ih)
+			if t, ok := client.Torrent(hash); ok {
+				lpdPeer(t, addr.String())
+				ignore[t] = true
+			}
+		}
+		// LPD is the only source of local IP's. So, add it to all active torrents.
+		for t := range client.torrents {
+			if _, ok := ignore[t]; ok {
+				continue
+			}
+			lpdPeer(t, addr.String())
+		}
+		mu.Unlock()
+	}
+}
+
+func (m *LPDConn) announcer(client *Client) {
+	var refresh time.Duration = 0
+	var next *Torrent
+	var queue []*Torrent
+
+	for {
+		mu.Lock()
+		m.force.Clear()
+		mu.Unlock()
+
+		//log.Println("LPD", refresh)
+
+		select {
+		case <-m.stop.LockedChan(&mu):
+			return
+		case <-m.force.LockedChan(&mu):
+		case <-time.After(refresh):
+		}
+
+		mu.Lock()
+		// add missing torrent to send queue
+		for t := range client.torrents {
+			if _, ok := lpdContains(queue, t); !ok {
+				queue = append(queue, t)
+			}
+		}
+
+		if next == nil {
+			if len(queue) > 0 {
+				next = queue[0]
+			}
+		}
+
+		// remove stopped torrent from queue
+		var remove []*Torrent
+		for _, t := range queue {
+			if _, ok := client.torrents[t]; !ok {
+				remove = append(remove, t)
+			}
+		}
+		for _, t := range remove {
+			if i, ok := lpdContains(queue, t); ok {
+				if next == t { // update next to next+1
+					n := i + 1
+					if n >= len(queue) {
+						next = nil
+					} else {
+						next = queue[n]
+					}
+				}
+				queue = append(queue[:i], queue[i+1:]...)
+			}
+		}
+		lpd.refresh()
+
+		var ihs string
+		var old []byte
+
+		port := client.LocalPort()
+		count := 0
+		for next != nil {
+			ihs += fmt.Sprintf(bep14_announce_infohash, strings.ToUpper(next.InfoHash().HexString()))
+			req := fmt.Sprintf(bep14_announce, m.host, strconv.Itoa(port), ihs)
+			buf := []byte(req)
+			if len(buf) >= 1400 {
+				break
+			}
+			old = buf
+			if i, ok := lpdContains(queue, next); ok {
+				i++
+				if i >= len(queue) {
+					next = nil
+				} else {
+					next = queue[i]
+				}
+			}
+			count++
+			if bep14_max > 0 && count >= bep14_max {
+				break
+			}
+		}
+		mu.Unlock()
+
+		if len(old) > 0 {
+			//log.Println("LPD", string(old), len(old))
+			_, err := m.mcPublisher.Write(old)
+			if err != nil {
+				log.Println("announcer", err)
+			}
+		}
+
+		refresh = bep14_short_timeout
+		if next == nil { // restart queue
+			refresh = bep14_long_timeout
+		}
+	}
+}
+
+func (m *LPDConn) Close() {
+	m.stop.Set()
+	if m.mcListener != nil {
+		m.mcListener.Close()
+		m.mcListener = nil
+	}
+}
+
+type LPDServer struct {
+	conn4 *LPDConn
+	conn6 *LPDConn
+
+	peers map[int64]string // active local peers
+}
+
+func lpdStart(client *Client) {
+	lpd = &LPDServer{}
+
+	lpd.peers = make(map[int64]string)
+
+	lpd.conn4 = lpdConnNew("udp4", bep14_host4)
+	if lpd.conn4 != nil {
+		go lpd.conn4.receiver(client)
+		go lpd.conn4.announcer(client)
+	}
+
+	lpd.conn6 = lpdConnNew("udp6", bep14_host6)
+	if lpd.conn6 != nil {
+		go lpd.conn6.receiver(client)
+		go lpd.conn6.announcer(client)
+	}
+}
+
+func (m *LPDServer) refresh() {
+	now := time.Now().UnixNano()
+	var remove []int64
+	for t := range m.peers {
+		// remove old peers who did not refresh for 2 * bep14_long_timeout
+		if t+(2*bep14_long_timeout).Nanoseconds() < now {
+			remove = append(remove, t)
+		}
+	}
+	for _, t := range remove {
+		delete(m.peers, t)
+	}
+}
+
+func (m *LPDServer) peer(peer string) {
+	now := time.Now().UnixNano()
+
+	var remove []int64
+	for k, v := range m.peers {
+		if v == peer {
+			remove = append(remove, k)
+		}
+	}
+
+	m.peers[now] = peer
+
+	for _, v := range remove {
+		delete(m.peers, v)
+	}
+}
+
+func lpdContains(queue []*Torrent, e *Torrent) (int, bool) {
+	for i, t := range queue {
+		if t == e {
+			return i, true
+		}
+	}
+	return -1, false
+}
+
+func lpdForce() {
+	if lpd.conn4 != nil {
+		lpd.conn4.force.Set()
+	}
+	if lpd.conn6 != nil {
+		lpd.conn6.force.Set()
+	}
+}
+
+func lpdStop() {
+	mu.Lock()
+	defer mu.Unlock()
+	if lpd != nil {
+		if lpd.conn4 != nil {
+			lpd.conn4.Close()
+			lpd.conn4 = nil
+		}
+		if lpd.conn6 != nil {
+			lpd.conn6.Close()
+			lpd.conn6 = nil
+		}
+		lpd = nil
+	}
+}
+
+func lpdPeers(t *Torrent) {
+	for _, p := range lpd.peers {
+		lpdPeer(t, p)
+	}
+}
+
+func lpdPeer(t *Torrent, p string) {
+	host, port, err := net.SplitHostPort(p)
+	if err != nil {
+		return
+	}
+	pi, err := strconv.Atoi(port)
+	if err != nil {
+		return
+	}
+	ip := net.ParseIP(host)
+	peer := PeerInfo{
+		Addr:   &net.UDPAddr{IP: ip, Port: pi},
+		Source: PeerSourceDhtGetPeers,
+	}
+	t.AddPeers([]PeerInfo{peer})
+}
--- a/bep14_test.go
+++ b/bep14_test.go
@@ -1,0 +1,33 @@
+package torrent
+
+import (
+	"bufio"
+	"bytes"
+	"net/http"
+	"testing"
+)
+
+func TestMultiInfohash(t *testing.T) {
+	html := `BT-SEARCH * HTTP/1.1
+Host: 239.192.152.143:6771
+Port: 3333
+Infohash: 123123
+Infohash: 2222
+cookie: name=value
+
+
+`
+	req, err := http.ReadRequest(bufio.NewReader(bytes.NewReader([]byte(html))))
+	if err != nil {
+		t.Error("receiver", err)
+		return
+	}
+	var ihs []string = req.Header[http.CanonicalHeaderKey("Infohash")]
+	if ihs == nil {
+		t.Error("receiver", "No Infohash")
+		return
+	}
+	for _, ih := range ihs {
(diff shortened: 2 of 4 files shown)


client.go
```diff
@@ -348,6 +348,8 @@ func NewClient(cfg *ClientConfig) (cl *Client, err error) {
 		},
 	}
 
+	lpdStart(cl)
+
```
philrhc, line 352: probably should have a configuration flag to enable peer discovery

peer.go
```diff
@@ -296,12 +297,13 @@ func (cn *Peer) writeStatus(w io.Writer) {
 		prioStr += ": " + err.Error()
 	}
 	fmt.Fprintf(w, "bep40-prio: %v\n", prioStr)
-	fmt.Fprintf(w, "last msg: %s, connected: %s, last helpful: %s, itime: %s, etime: %s\n",
+	fmt.Fprintf(w, "last msg: %s, connected: %s, last helpful: %s, itime: %s, etime: %s, discovery source: %s\n",
```
philrhc, line 300: possibly not the right place to add this

bep14.go
```diff
@@ -0,0 +1,369 @@
...
+	m.mcListener, err = net.ListenMulticastUDP(m.network, nil, m.addr)
+	if err != nil {
+		log.Println("LPD unable to start", err)
+		return nil
+	}
+	m.mcPublisher, err = net.DialUDP(network, nil, m.addr)
+	if err != nil {
+		fmt.Println("Error dialing UDP:", err)
+		return nil
+	}
```
philrhc, line 73: this whole file is almost identical to libtorrent except that i couldn't run multiple torrent instances on the same machine to test unless this is added

philrhc:
happy to add more testing here, i've been testing in a scratch app

bep14.go
```diff
@@ -0,0 +1,373 @@
...
+	"strings"
+	"time"
+
+	"github.com/anacrolix/missinggo"
+	"github.com/anacrolix/sync"
+	"github.com/anacrolix/torrent/metainfo"
+)
+
+var lpd *LPDServer
+var mu sync.Mutex
```
philrhc, line 20: i think this needs to be changed...

anacrolix:
Are you interested in refactoring it a bit (totally fine if not). I think we could move the LPD stuff into a separate package, probably just `./lpd`, or `bep14` or `lsd` (not sure why LSD and LPD are mixed and matched in the literature here), mainly just to not clutter the API more.

What are the licence requirements for this? I note @axet has forgotten the MPL2 licence throughout his codebase. He's got LGPL3 for his code, so does that need to be reflected here?

bep14.go
anacrolix: Yeah makes sense. libtorrent was aiming for a singleton instance for mobile devices IIRC.

bep14.go
anacrolix: It'll need checking. Global locks aren't bad if they're used correctly and don't matter near as much as people think.

bep14.go
```diff
@@ -0,0 +1,369 @@
...
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/anacrolix/missinggo"
+	"github.com/anacrolix/sync"
+	"github.com/anacrolix/torrent/metainfo"
+)
+
+var lpd *LPDServer
```
anacrolix, line 19: This particular global probably shouldn't exist per your other comment.

client.go
anacrolix: LocalServiceDiscovery bool, in ClientConfig. Probably default to false for security reasons but I'm open.

peer.go
anacrolix: This should already exist, see PeerConn.connectionFlags.

anacrolix:
Thanks :)

anacrolix:
There are some race conditions. Run the tests with -race on your system.

philrhc:
>There are some race conditions. Run the tests with -race on your system.

thank you, this has been an invaluable tool.

>Are you interested in refactoring it a bit (totally fine if not). I think we could move the LPD stuff into a separate package, probably just ./lpd, or bep14 or lsd (not sure why LSD and LPD are mixed and matched in the literature here), mainly just to not clutter the API more

would like to leave the refactoring question until last, to see whether i have enough time to contribute the refactoring


-- 

only the test TestHolepunchConnect now fails (with LPD enabled by default, to catch as many errors as possible). It seems that this is because the peer is added by local peer discovery before `waitForConns` starts waiting here: `t.cl.event.Wait()`

bep14.go
philrhc: i've resolved this by adding lpd to the client

bep14.go
philrhc: mu is no longer a global variable, it has been added to the LPDServer

philrhc:
well i have fixed TestHolepunchConnect with https://github.com/anacrolix/torrent/pull/998/commits/24a8f65f859849c1987effef26b1830f06f9e67c

please can you take a look at this in particular? is this how you would expect this to look? https://github.com/anacrolix/torrent/pull/998/commits/24a8f65f859849c1987effef26b1830f06f9e67c

philrhc:
i've added TestDiscovery, it's a little crude https://github.com/anacrolix/torrent/pull/998/commits/76089ef27308938e96ca922f50035a00ca11364c

philrhc:
it looks like they repeatedly add themselves and each other as peers, i will work on preventing this at the lpd level

bep14.go
```diff
@@ -0,0 +1,378 @@
...
+	if lpd.conn4 != nil {
+		go lpd.conn4.receiver(client)
+		go lpd.conn4.announcer(client)
+	}
+
+	lpd.conn6 = lpdConnNew("udp6", bep14_host6, lpd)
+	if lpd.conn6 != nil {
+		go lpd.conn6.receiver(client)
+		go lpd.conn6.announcer(client)
+	}
```
philrhc, line 279: is it required to flag whether one or both ip4/6 are used?

bep14.go
```diff
@@ -0,0 +1,378 @@
...
+			continue
+		}
+
+		port := req.Header.Get("Port")
+		if port == "" {
+			log.Println("receiver", "No port")
+			continue
+		}
+
+		addr, err := net.ResolveUDPAddr(m.network, net.JoinHostPort(from.IP.String(), port))
```
philrhc, line 117: welcoming any suggestions on how to understand if self was the announcer here,
only using client.LocalPort seems a little fragile

bep14.go
```diff
@@ -0,0 +1,396 @@
...
+			log.Println("receiver", err)
+			continue
+		}
+
+		client.rLock()
+		if client.LocalPort() == addr.Port {
+			log.Println("discovered self")
+			client.rUnlock()
+			continue
+		}
```
philrhc, line 137: self is no longer discovered but only considering port number for the first listener that has one seems a bit fragile...
could publish peer id in the message but i guess this could cause parsing problems with other clients? https://bittorrent.org/beps/bep_0014.html

bep14_test.go
```diff
@@ -0,0 +1,80 @@
...
+	defer os.RemoveAll(greetingTempDir)
+
+	seederTorrent, _, _ := client1.AddTorrentSpec(TorrentSpecFromMetaInfo(mi))
+	leecherGreeting, _, _ := client2.AddTorrentSpec(func() (ret *TorrentSpec) {
+		ret = TorrentSpecFromMetaInfo(mi)
+		ret.ChunkSize = 2
+		return
+	}())
+
+	time.Sleep(2 * time.Second)
```
philrhc, line 62: would appreciate an alternative to the 2 second sleep

client.go
```diff
@@ -455,6 +462,9 @@ func (cl *Client) eachDhtServer(f func(DhtServer)) {
 
 // Stops the client. All connections to peers are closed and all activity will come to a halt.
 func (cl *Client) Close() (errs []error) {
+	if (cl.lpd != nil) {
```
philrhc, line 465: is this and the other nil checks how you would expect this to look?

bep14.go
```diff
@@ -0,0 +1,396 @@
...
+	m.mcListener, err = net.ListenMulticastUDP(m.network, nil, m.addr)
+	if err != nil {
+		log.Println("LPD unable to start", err)
+		return nil
+	}
+	m.mcPublisher, err = net.DialUDP(network, nil, m.addr)
+	if err != nil {
+		fmt.Println("Error dialing UDP:", err)
+		return nil
+	}
```
philrhc, line 72: i need an optional way to specify the desired network interface on which to multicast, i propose adding this to ClientConfig

anacrolix:
This is looking very good, thanks for taking feedback on board. I'll give a more detailed response shortly.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/anacrolix/torrent
 
-go 1.23
+go 1.23.0
+
+toolchain go1.24.1
```
philrhc, line 5: a little confused about how i've caused this, i think it's because of requiring:
"golang.org/x/net/ipv4"
"golang.org/x/net/ipv6"
but i can't work out exactly why...

bep14_test.go
```diff
@@ -0,0 +1,80 @@
...
+func waitForPeers(t *Torrent) {
+	t.cl.lock()
+	defer t.cl.unlock()
+	for {
+		if t.numTotalPeers() > 0 {
+			return
+		}
+		t.cl.event.Wait()
+	}
+}
```
philrhc, line 80: TestHolePunch contains this code, do you know the difference and why this might work and t.conns not?:

```
func waitForConns(t *Torrent) {
	t.cl.lock()
	defer t.cl.unlock()
	for {
		for range t.conns {
			return
		}
		t.cl.event.Wait()
	}
}
```

bep14.go
```diff
@@ -0,0 +1,421 @@
...
+	m.host = host
+
+	var err error
+
+	m.addr, err = net.ResolveUDPAddr(m.network, m.host)
+	if err != nil {
+		log.Println("LPD unable to start", err)
+		return nil
+	}
+	m.mcListener, err = net.ListenMulticastUDP(m.network, nil, m.addr)
```
philrhc, line 89: should only listen on the interfaces which are multicasting?

go.mod
anacrolix: There's a bug in Go, they just pushed a fix to master https://github.com/golang/go/issues/65847#issuecomment-2657948532

bep14_test.go
philrhc: resolved https://github.com/anacrolix/torrent/pull/998/commits/fbd780ca9a23c4185172ad947fa0fc973104fe3b

philrhc:
okay, i'm happy with it now, it will wait for your review

> Are you interested in refactoring it a bit (totally fine if not). I think we could move the LPD stuff into a separate package, probably just ./lpd, or bep14 or lsd (not sure why LSD and LPD are mixed and matched in the literature here), mainly just to not clutter the API more.

i spent quite a lot longer resolving race conditions than i thought, so would appreciate it if you can take the refactoring forward.

bep14_test.go
```diff
@@ -0,0 +1,80 @@
...
+		return
+	}())
+
+	time.Sleep(2 * time.Second)
+	waitForPeers(seederTorrent)
+	require.Equal(t, seederTorrent.numTotalPeers(), 2)
+	require.Equal(t, len(client1.lpd.peers), 2)
+	waitForPeers(leecherGreeting)
+	require.Equal(t, leecherGreeting.numTotalPeers(), 2)
+	require.Equal(t, len(client2.lpd.peers), 2)
```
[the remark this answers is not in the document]
philrhc: i resolved this by adding a config setting to avoid using IPV6, so only 1 peer should be discovered
https://github.com/anacrolix/torrent/pull/998/commits/fbd780ca9a23c4185172ad947fa0fc973104fe3b

bep14.go
```diff
@@ -0,0 +1,476 @@
...
+
+	"github.com/anacrolix/missinggo"
+	"github.com/anacrolix/torrent/metainfo"
+)
+
+// http://bittorrent.org/beps/bep_0014.html
+// TODO http://bittorrent.org/beps/bep_0026.html
+
+const (
+	bep14_host4    = "239.192.152.143:6771"
```
anacrolix, line 26: Camel case is the convention but no stress. 

Props if you're an old C programmer

bep14.go
```diff
@@ -0,0 +1,476 @@
...
+		}
+	}
+	if m.network == "upd6" {
+		p := ipv6.NewPacketConn(m.mcPublisher)
+		if err := p.SetMulticastInterface(iface); err != nil {
+			log.Printf("Set multicast interface error: %v\n", err)
+			return err
+		}
+	}
+	return nil
```
anacrolix, line 68: Maybe return an error for unexpected network?

bep14.go
```diff
@@ -0,0 +1,476 @@
...
+func (m *LPDConn) receiver(client *Client) {
+	for {
+		m.lpd.mu.RLock()
+		if m.closed {
+			m.lpd.mu.RUnlock()
+			return
+		}
+		m.lpd.mu.RUnlock()
+
+		buf := make([]byte, 2000)
```
anacrolix, line 174: Where does this number come from? MTU?

bep14.go
```diff
@@ -0,0 +1,476 @@
...
+		_, from, err := m.mcListener.ReadFromUDP(buf)
+		if err != nil {
+			log.Println("receiver", err)
+			continue
+		}
+
+		req, err := http.ReadRequest(bufio.NewReader(bytes.NewReader(buf)))
+		if err != nil {
+			log.Println("receiver", err)
```
anacrolix, line 183: There is a defacto logging library in this repo. This message is going to be very hard to track down

bep14.go
```diff
@@ -0,0 +1,476 @@
...
+		}
+
+		refresh = bep14_short_timeout
+		if next == nil { // restart queue
+			refresh = bep14_long_timeout
+		}
+	}
+}
+
+type LPDServer struct {
```
anacrolix, line 354: Can this be lower cased to avoid exposing it as public API? Convention is lpdServer

bep14_test.go
```diff
@@ -0,0 +1,79 @@
...
+	client1, err := NewClient(config)
+	require.NoError(t, err)
+	defer client1.Close()
+	
+	client2, err := NewClient(config)
+	require.NoError(t, err)
+	defer client2.Close()
+
+	greetingTempDir, mi := testutil.GreetingTestTorrent()
+	defer os.RemoveAll(greetingTempDir)
```
anacrolix, line 52: This might be copypasta from my old tests but T. Cleanup should be here somewhere

client.go
```diff
@@ -348,6 +350,11 @@ func NewClient(cfg *ClientConfig) (cl *Client, err error) {
 		},
 	}
 
+	if cfg.LocalServiceDiscovery.Enabled {
```
anacrolix, line 353: I haven't checked this elsewhere, but Enabled internal fields are always suspect to me. Can the config not be optional and/or an enabled field be external?

client.go
anacrolix: Yeah, there's also a close function slice you can put your close handler into. That might be preferred. 

The syntax looks odd, there's extra parentheses. Try running go fmt ./... on your PR

client.go
```diff
@@ -1416,14 +1426,15 @@ func (cl *Client) AddTorrentInfoHashWithStorage(
 func (cl *Client) AddTorrentOpt(opts AddTorrentOpts) (t *Torrent, new bool) {
 	infoHash := opts.InfoHash
 	cl.lock()
-	defer cl.unlock()
 	t, ok := cl.torrentsByShortHash[infoHash]
 	if ok {
+		cl.unlock()
```
anacrolix, line 1431: I wonder why you removed the defer. Although I think it's a good call. I have learned the hard way that panic safety is an anti pattern in Go

config.go
```diff
@@ -45,6 +45,13 @@ type ClientDhtConfig struct {
 	DHTOnQuery func(query *krpc.Msg, source net.Addr) (propagate bool)
 }
 
+type LocalServiceDiscoveryConfig struct {
+	Enabled bool
```
anacrolix, line 49: This disallows replacement with alternatives, and requires this code and its caller track enabled state. Best to just create and configure the insurance externally and assume it's enabled internally if the instance receives any calls.

anacrolix (commented):
Mostly just a few questions and stylistic suggestions that are optional. 

I'll give it a a spin and add any necessary cleanup to fit in to any tests I run unless you beat me to it. 

Was there any result on the licence question?

Thanks

anacrolix:
> okay, i'm happy with it now, it will wait for your review
> 
> > Are you interested in refactoring it a bit (totally fine if not). I think we could move the LPD stuff into a separate package, probably just ./lpd, or bep14 or lsd (not sure why LSD and LPD are mixed and matched in the literature here), mainly just to not clutter the API more.
> 
> i spent quite a lot longer resolving race conditions than i thought, so would appreciate it if you can take the refactoring forward.

Ok I see this, I'll fix anything that comes up in my testing myself.

anacrolix:
I'll rebase or merge the CI fixes from master.

bep14_test.go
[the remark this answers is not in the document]
anacrolix: Yeah I think I've had to do this elsewhere. 

I originally used dual stack everywhere but that actually causes problems due to Go IP handling being not great. See my cranky Lobsters comment.

anacrolix:
This is going to need me to test it properly. I'll keep it open and plod away on it.

Others are welcome to try it out and review.

philrhc:
thanks for the comments, i'm happy for this to plod away because i am testing/prototyping with it on my branch, so no rush

will resolve some of the comments when i get some time, thanks again

anacrolix:
I think this needs a standalone cmd for testing. The concurrency and way it hooks into the main client is very old and crufty.

philrhc:
happy to take some/all of this on if you provide more detailed instructions, i am new to production go code and would like to learn, thanks

anacrolix:
I assume a lot of it was taken from downstream. It's not in a great style, and doesn't do things in the preferred way. That wouldn't be a problem if it passed the tests reliably which I think it didn't when I tested. So to take this code I'll at least need all the tests to pass reliably.

A cmd that allows listening and sending the discovery messages from the command line will help a lot. It could be added to cmd/torrent or a new cmd. It should allow interactive with LPD without requiring a client to be run.

bep14_test.go
philrhc: https://github.com/anacrolix/torrent/pull/998/commits/38813a25846a4bcb9e0d7709821eb37c5b73d1bb

bep14.go
philrhc: https://github.com/anacrolix/torrent/pull/998/commits/48480e14cbc34b64164d7399bc52f626ff8fe72c

bep14.go
philrhc: hmm, i haven't found an elegant way to do this, without using `m.network != "udp4" && m.network != "udp6"`

now that you draw my attention to it, i'm curious why the p variable is not garbage collected... i wanted to add it to the lpdConn struct but it has type `*ipv4.PacketConn` or `*ipv6.PacketConn`

what do you think?

bep14.go
philrhc: LPDServer is used in a few places in client.go 

happy for any suggestions how to improve this

philrhc:
it seems that the flaky test is fixed with this change https://github.com/anacrolix/torrent/pull/998/commits/65f4322796a5e3862419d6630a19825b5e991d5c

i think the windows runner is very slow, the test is timing out before it has a chance to start, have you seen a problem like this before?
https://github.com/anacrolix/torrent/actions/runs/14644323699/job/41094412038

philrhc:
i've been using `sudo tcpdump -i any udp port 6771 -A` to listen to the discovery messages, here is the log when running TestDiscovery

```
sudo tcpdump -i any udp port 6771 -A

08:43:24.934997 enxac91a1741908 Out IP bsccs374.int.bsc.es.50386 > 239.192.152.143.6771: UDP, length 119
E.....@...T.TX2 .......s...BBT-SEARCH * HTTP/1.1
Host: 239.192.152.143:6771
Port: 34453
Infohash: 530392E356A40161C17DBBFEF4DB07AD029F73AF

08:43:24.935254 enxac91a1741908 Out IP bsccs374.int.bsc.es.49073 > 239.192.152.143.6771: UDP, length 119
E....f@...bCTX2 .......s...BBT-SEARCH * HTTP/1.1
Host: 239.192.152.143:6771
Port: 43793
Infohash: 530392E356A40161C17DBBFEF4DB07AD029F73AF
```

config.go
philrhc: i have addressed this here, please let me know what you think
https://github.com/anacrolix/torrent/pull/998/commits/4b0e192ab7a65537a7d719024a001fd8228e65e5

philrhc:
i have found an 

> it seems that the flaky test is fixed with this change [65f4322](https://github.com/anacrolix/torrent/commit/65f4322796a5e3862419d6630a19825b5e991d5c)
> 
> i think the windows runner is very slow, the test is timing out before it has a chance to start, have you seen a problem like this before? https://github.com/anacrolix/torrent/actions/runs/14644323699/job/41094412038

i have found an issue tracking the windows runner speed https://github.com/actions/runner-images/issues/7320

client.go
philrhc: i think you have resolved this by adding the close handler to the close fn slice

client.go
philrhc: i unlock the client because when t.go requires the lock LPD adds peers to a torrent
https://github.com/anacrolix/torrent/blob/589b1ebd9f3b8989537178ad4c762b49df64af4d/t.go#L245

go.mod
philrhc: i am hypothesizing that this causes the slowness seen here https://github.com/anacrolix/torrent/pull/998#issuecomment-2841017921

that two versions of Go are being downloaded

philrhc: yes i see something funky happening here
https://github.com/philrhc/torrent/actions/runs/14750112335/job/41406098250
Common Go seems to download 3 versions of the toolchain, 1.23.0, 1.23.8 and the upgraded 1.24.2. I wonder if this will happen to any pull request that upgrades the go version. I will investigate how this action cache works...

philrhc: ok, i think the caching is working okay now that all versions of go are 1.24.2. I am speculating that some part of the code does not work with Windows and this causes the test timeout...

philrhc:
i am currently not able to find the reason the windows runner cannot complete TestDiscovery. I'll give this a rest now...

bep14.go
anacrolix: You shouldn't rely on GC to clean up resources in Go. In most code depending on stdlib, os.File has a special handler to clean up because people forget so often. I expect the ipv4/6 PacketConn aren't doing this underneath or something.

Try to put those into a common interface type? Or you can make an interface { a | b } type to hold it possibly.

anacrolix:
> i am currently not able to find the reason the windows runner cannot complete TestDiscovery. I'll give this a rest now...

Ah, Windows probably does very different things with regard to multicast. I had to do similar hackery for anacrolix/dms. https://github.com/anacrolix/dms/blob/1291628af96c12c6a81ad900021f9da912bc8fd5/ssdp/ssdp.go#L115

philrhc:
> > i am currently not able to find the reason the windows runner cannot complete TestDiscovery. I'll give this a rest now...
> 
> Ah, Windows probably does very different things with regard to multicast. I had to do similar hackery for anacrolix/dms. https://github.com/anacrolix/dms/blob/1291628af96c12c6a81ad900021f9da912bc8fd5/ssdp/ssdp.go#L115

good to know, thanks. I'm planning to take a look at this again soon

philrhc:
hi, thought i'd check in :)
i'm testing this on the Fabric testbed, and so far so good, when it scales to 30 nodes then sometimes one node doesn't complete the download, it doesn't seem to be a network issue because the failure can change over repeated tests with the same set of nodes.
i got a bit sidetracked optimising sqlite, the confluence sqlite options have resulted in very slow downloads. https://github.com/anacrolix/confluence/blob/dc1134d3b74b92a161809ccb2694fad80d877b01/main.go#L168

 thank you for the fantastic work over the years, it's been a pleasure to use torrent and confluence :)

PR #1010, opened 2025-05-14
Support announce completed when torrent download finish

charleshuang3:
It is a early PR. @anacrolix please teach me:

- where is the best place to add tests
- where is the best place to receive the channel close signal


tracker_scraper.go
```diff
@@ -265,15 +267,30 @@ func (me *trackerScraper) Run() {
...
-		select {
-		case <-me.stopCh:
-			return
-		case <-me.t.closed.Done():
-			return
-		case <-reconsider:
-			// Recalculate the interval.
-			goto recalculate
-		case <-time.After(time.Until(ar.Completed.Add(interval))):
+		if !me.complete {
```
charleshuang3, line 270: without this, `case <-me.t.complete.On()` keep triggered

anacrolix:
The tracker code was rewritten, this PR would no longer apply. Thanks!

anacrolix:
Completion events were implemented in a recent version of anacrolix/torrent, see the changelog.

Issue #175, opened 2017-08-09
Adopt bittorrent 2 spec


ewhal:
https://github.com/arvidn/libtorrent/issues/2197
https://github.com/atomashpolskiy/bt/issues/28

anacrolix:
![But why?](https://media.giphy.com/media/1M9fmo1WAFVK0/giphy.gif)

christian-roggia:
Here you will find a good reddit TL;DR about the changes and improvements of **The BitTorrent Protocol Specification v2**:
https://www.reddit.com/r/programming/comments/6safyl/the_bittorrent_protocol_specification_v2/

For more detailed information read the bittorrent.org BEP:
http://bittorrent.org/beps/bep_0052.html

The most important improvements are:

- Introduction of SHA256 and globally unique hashes that will prevent hash collisions and therefore all known hash collision attacks
- Merkle tree for hashing pieces instead of chunk hashes


Not really sure this is currently really needed, but it is definitely something soon or later should be fully implemented (with the help of the community).

anacrolix:
A useful link from a related issue: https://blog.libtorrent.org/2020/09/bittorrent-v2/.

gitpubber:
@anacrolix when will you support this?

anacrolix:
Not until I see some uptake in the wild. Currently it appears to be zero.

KyleSanderson:
Any update on this one? This is absolutely "in the wild".

anacrolix:
I am considering this, there's been an uptick in the community, I think usage might start to appear.

Is there any interest in funding this support?

KyleSanderson:
> I am considering this, there's been an uptick in the community, I think usage might start to appear.
> 
> Is there any interest in funding this support?

Specifically I have a PR into a project that already uses this library. https://github.com/autobrr/autobrr/pull/491 

This breaks a handful of sites by parsing the contents as the clients well support V2, but the library does not. As V2 has become widely accepted and mainstream over the last year, this is only going to become more prevalent as time progresses.

So, similar situation to you, as a user there's pain. I have an old parser for v1 I can dust off, but then the maintenance burden shifts.

kovalensky:
When? Tell me honey when? I'm just to hungry for this, give me, give me that juicy BitTorrent v2 support, make it faster, deeper in implementation details, make it swim inside my web server, take all of my resources, make it bigger.

[...]

PR #1011, opened 2025-05-16
Small fixes for v2 support

Jipok:


anacrolix:
Thanks I'll verify soon

Jipok:
But I still couldn't get pure v2 to work. Here's my code for creating a torrent:
https://gist.github.com/Jipok/98cb92486378ad04b6c35e8bfe106314
It seems to create a valid torrent. [Another client](https://gist.github.com/Jipok/78851ccddaccc884a6b569aed2ae6e03) finds it by infohash, gets the metadata correctly and then gets stuck at 0.8%. There's no information in the logs.

anacrolix:
Sorry for the delay, haven't had time to look yet. Have you tried with some regular, pure-v2 torrents from the wild?

anacrolix:
Without looking too carefully I'll guess the v2 you're generating either isn't valid/correct, or any code that performs this function for you inside anacrolix/torrent doesn't do it right. I've tested only really only the reference torrents provided with the v2 spec. See https://github.com/anacrolix/torrent/issues/175#issuecomment-1987015148 for the current state of the implementation.

anacrolix:
I'll close this, I think there's not enough value and it's not clear what is required now.

PR #1015, opened 2025-07-24
stop tracker when err occcurs

ucwong:


anacrolix:
I don't think this is correct. The error is put into lastAnnounce, which you can sample from the client status.

ucwong:
> I don't think this is correct. The error is put into lastAnnounce, which you can sample from the client status.

Does it mean that the tracker with error won't be running again during the whole life cycle of the current torrent?

anacrolix:
Yes

PR #1024, opened 2025-10-18
unswitch cl.forwardPort

milahu:
avoid calling this function
when `cl.config.NoDefaultPortForwarding` is false

https://github.com/anacrolix/torrent/blob/a6f1f60a7af4558ac05779ea9581305523354604/portfwd.go#L37-L42

also add comments with config alias names

- `cfg.NoDefaultPortForwarding` aka `cfg.DisableUpnp`
- `cfg.NoDHT` aka `cfg.DisableDHT`

### DIFF
--- a/config.go
+++ b/config.go
@@ -40,6 +40,7 @@ type ClientTrackerConfig struct {
 
 type ClientDhtConfig struct {
 	// Don't create a DHT.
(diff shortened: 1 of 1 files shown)


anacrolix:
Thanks!

PR #1025, opened 2025-10-19
add HTTP storage backend storage.HttpStorageImpl

milahu:
read metadata and content files from a remote HTTP server

write support is not planned

use case:
[btcache-go](https://github.com/milahu/btcache-go) - A caching reverse-proxy BitTorrent client, written in Go

first i tried to implement this by monkeypatching the default storage methods
like `st.TorrentImpl.Piece = func(p metainfo.Piece) storage.PieceImpl {`
see https://github.com/milahu/btcache-go/commit/188d737edbedf86ab5bedf9063a631b4d6be071a

but that failed, because many symbols are private
(so i would have to copy and rename lots of code)
and because some structs are embedded by value, not by reference
see https://github.com/milahu/anacrolix-torrent/commit/7cb993a31e38b5bc9f05f206b1cd11e5c0840b54

so it was much easier to copy
storage/file-client.go
storage/file-piece.go
storage/file-torrent.go
storage/file-torrent-io.go
to
storage/http-storage.go
storage/http-piece.go
storage/http-torrent.go
storage/http-torrent-io.go
... and go from there (ha! "go"...)

the biggest change is in storage/http-torrent-io.go
where the "fetch and cache pieces" logic is implemented
in `func (tio HttpTorrentImplIO) ReadAt`

a related issue is https://github.com/jpillora/cloud-torrent/issues/24
but it mainly deals with writing to remote storage backends
while i am only interested in reading from remote storage backends

too messy to merge for now
but maybe someone can use it : )

### DIFF
--- a/client.go
+++ b/client.go
@@ -51,6 +51,7 @@ import (
 	"github.com/anacrolix/torrent/storage"
 	"github.com/anacrolix/torrent/types/infohash"
 	infohash_v2 "github.com/anacrolix/torrent/types/infohash-v2"
+	"github.com/anacrolix/torrent/version"
 )
 
 const webseedRequestUpdateTimerInterval = 5 * time.Second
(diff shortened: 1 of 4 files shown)


milahu:
abandoned for [btcache-py](https://github.com/milahu/btcache-py)
which uses a simpler strategy (default file storage + hidden peers)

> too messy to merge for now
> but maybe someone can use it : )

not usable in its current state

it works in a trivial test when all clients run on localhost
but it fails on a real-world test between different machines
where http requests fail with `context deadline exceeded`

another problem here is `t.SetHaveAllPieces()`
because peers can block our client for faking our "haves"

Issue #444, opened 2020-12-09
Retain peers and tracker intervals between torrent "sessions"


anacrolix:
Per https://github.com/anacrolix/torrent/issues/413#issuecomment-741792411, tracker intervals, and peers are not retained across Client (and more specifically Torrent) restarts. In some situations this can be very counterproductive.

PR #1028, opened 2025-11-09
Feat/persist session state #444

ljluestc:
Persist per-torrent session state to `DataDir/.session/<infohash>.json`:
  - Peers (host:port) for fast bootstrap after restart.
  - Trackers: last-announce `completed_unix` and `interval_seconds` to honor backoff.
- Load peers on torrent creation; save peers on torrent close.
- Initialize each `trackerScraper` from persisted last-announce and delay first announce until the saved interval elapses.
- Persist tracker announce state after each announce.
- Fix deadlock on startup: load persisted peers under the Client lock using `t.addPeers`.

Implementation highlights
- `session_persistence.go`: JSON read/write helpers; peer/announce state translation.
- `client.go`: load peers during torrent init; save on close; deadlock-safe peer load.
- `torrent.go`: initialize tracker announcers from persisted state.
- `tracker_scraper.go`: honor/persist announce interval.
- Tests: `TestPersistPeersAcrossSessions`, `TestTrackerIntervalRetainedBetweenSessions` (uses `httptest.Server`; discovery listeners disabled to keep tests isolated).

Backwards compatibility
- Disabled when `DataDir` is empty.
- No migrations; small per-torrent JSON.

Performance and safety
- Small writes on close/announce.
- Peer list capped to 200.
- Local plaintext under `DataDir`.

### DIFF
--- a/session_persistence.go
+++ b/session_persistence.go
@@ -1,0 +1,213 @@
+package torrent
+
+import (
+	"encoding/json"
+	"net"
+	"os"
+	"path/filepath"
+	"strconv"
+	"sync"
+	"time"
+)
+
+// Session persistence of lightweight torrent state across client restarts.
+// We persist a small set of data:
+// - Reserve peers (addresses) to bootstrap connections without waiting for DHT/trackers.
+// - Last tracker announce completion time and interval per tracker URL to honor backoffs.
+
+type persistedTrackerState struct {
+	URL             string `json:"url"`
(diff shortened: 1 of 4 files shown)


anacrolix:
I don't think peer persistence is something I want to pursue in the library at the moment. However I definitely support third-party implementations doing this, so if you need something exposed/exported to make it possible to do cleanly, please ask/contribute that.

PR #1033, opened 2025-11-27
Revert "Use fixed Option.Iter"

ucwong:
This reverts commit 5dc10c807690777b49f0264e0039c008a2440854.

```
./client-tracker-announcer.go:344:18: cannot range over indexed.FirstInRange(me.overdueIndex, start, end).Iter (value of type func() "iter".Seq[nextAnnounceRecord]): func must be func(yield func(...) bool): wrong argument count
FAIL	github.com/anacrolix/torrent/storage [build failed]
```


anacrolix:
Thanks, I fixed the import instead. I don't want to get stuck with a bad Option.Iter.

PR #1040, opened 2026-02-22
fix: data race in BenchmarkConnectionMainReadLoop

Giulio2002:
## Problem

`BenchmarkConnectionMainReadLoop` fails with a data race on all CI platforms (stable/ubuntu, stable/macos, oldstable/ubuntu):

```
Previous write at 0x00c0001b2958 by goroutine 13499:
  github.com/RoaringBitmap/roaring.(*roaringArray).insertNewKeyValueAt()
  github.com/RoaringBitmap/roaring.(*Bitmap).Add()
  github.com/anacrolix/torrent.BenchmarkConnectionMainReadLoop()
      peerconn_test.go:123

Goroutine 13541 (running) created at:
  time.goFunc()
```

## Cause

`t._pendingPieces.Add(0)` at line 123 is called without holding the client lock, racing with timer goroutines spawned by the torrent client internals that also access the roaring bitmap.

## Fix

Wrap `_pendingPieces.Add(0)` with `cl.lock()`/`cl.unlock()` to match the locking discipline used elsewhere in the benchmark (e.g. `pendAllChunkSpecs` is called inside `cl.lock()`).

### DIFF
--- a/webtorrent/tracker-client.go
+++ b/webtorrent/tracker-client.go
@@ -308,17 +308,18 @@ func (tc *TrackerClient) announce(event tracker.AnnounceEvent, infoHash [20]byte
 	}
 
 	tc.mu.Lock()
-	defer tc.mu.Unlock()
 	err = tc.writeMessage(data)
 	if err != nil {
+		tc.mu.Unlock()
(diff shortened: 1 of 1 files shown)


Giulio2002:
This just *should fix your CI. you are welcome

PR #1041, opened 2026-03-09
migration to roaring v2 and used GetCardinalityInRange API

DhruvilK7:


anacrolix:
I've incorporated this into master, and retained your authorship on the commit that kicked it off. Thank you!

PR #1046, opened 2026-04-15
Fix Reader

skunkie:
Fix race condition in reader context synchronization
The reader struct had a data race where the ctx field was accessed without proper synchronization, causing potential panics and inconsistent behavior in concurrent scenarios.

Changes
reader.go:85-87: Added mutex lock/unlock in SetContext before writing r.ctx
reader.go:171-173: Added mutex lock/unlock in read before reading r.ctx
reader.go:178-181: Added mutex lock/unlock in ReadContext before writing r.ctx

Testing
All existing tests pass with -race flag, including:
TestReceiveChunkStorageFailureSeederFastExtensionDisabled
TestReceiveChunkStorageFailure

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -83,7 +83,9 @@ type reader struct {
 }
 
 func (r *reader) SetContext(ctx context.Context) {
+	r.mu.Lock()
 	r.ctx = ctx
+	r.mu.Unlock()
 }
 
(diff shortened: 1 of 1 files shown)


skunkie:
Maybe this PR is worth a look.

anacrolix:
Not sure how I missed this, thx.

PR #1047, opened 2026-04-21
Adds bep_0014 (local peer discovery)

philrhc:
Supercedes https://github.com/anacrolix/torrent/pull/998. I wanted to do a substantial refactor to address this comment: https://github.com/anacrolix/torrent/pull/998#issuecomment-2771234450

Upstreams BEP14 from https://gitlab.com/axet/libtorrent/-/blob/master/lpd.go

### DIFF
--- a/bep14.go
+++ b/bep14.go
@@ -61,31 +61,32 @@ type lpdConn struct {
 	cancel context.CancelFunc
 	force  chan struct{} // buffered(1): signal an immediate re-announce
 
-	lpd         *lpdServer
-	network     string // "udp4" or "udp6"
-	addr        *net.UDPAddr
-	mcListener  *net.UDPConn
-	mcPublisher *net.UDPConn
-	host        string // bep14Host4 or bep14Host6
-	logger      log.Logger
+	lpd          *lpdServer
+	network      string // "udp4" or "udp6"
+	addr         *net.UDPAddr
+	mcListener   *net.UDPConn
+	mcPublisher  *net.UDPConn
+	mcPacketConn mcPacketConn // non-nil when an explicit interface is bound
+	host         string       // bep14Host4 or bep14Host6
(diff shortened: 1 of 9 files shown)


bep14.go
```diff
@@ -0,0 +1,437 @@
...
+	"time"
+
+	"github.com/anacrolix/log"
+	"golang.org/x/net/ipv4"
+	"golang.org/x/net/ipv6"
+)
+
+// http://bittorrent.org/beps/bep_0014.html
+// TODO http://bittorrent.org/beps/bep_0026.html
+
```
philrhc, line 27: I wanted to pick up the thread from #998 where you flagged the LGPL-3 on axet's original lpd.go (https://gitlab.com/axet/libtorrent/-/blob/master/lpd.go).

The current bep14.go was initially adapted from that file but has since been substantially reworked — LPD goroutines decoupled from *Client behind an lpdClient interface, announce queue with rotation and packet fragmentation, plus unit and integration test coverage. Attribution to axet is preserved in the header comment.

Since this repo is MPL-2.0, I see a few possible paths:

  1. Keep bep14.go MPL-2.0 on the basis that the current code is sufficiently rewritten to no longer be a derivative work, with the axet attribution as a courtesy.
  2. Mark just bep14.go as LGPL-3 via an SPDX header — MPL is file-level copyleft, so a single LGPL file inside an MPL repo is legally clean.

Any preference? Happy to go with whatever you'd like, or a fourth option I haven't thought of.

anacrolix:
This looks good. I will decide shortly.

bep14.go
anacrolix: Let me do a few checks and decide.

anacrolix:
Alexey (axet) kindly provided his permission to licence derived files. I've pushed that to the PR.

PR #1051, opened 2026-05-07
Feat/windows storage fast stable

Aodi:
This PR improves storage configurability, download-path performance, and tracker recovery behavior, with a focus on making embedded usage more predictable and large active downloads more efficient and stable.

The changes fall into three main areas:

1. Improve file-storage performance on the classic path
2. Reduce per-piece completion persistence overhead with buffered checkpoints
3. Fix tracker announce state so removed and re-added torrents can rediscover peers correctly
4. Make default file I/O selection configurable at the right time for embedding applications
## Motivation
These changes were driven by real integration and runtime behavior observed in an application embedding this library:

- unnecessary filesystem overhead during active downloads on the classic file-storage path
- excessive per-piece durability work when using persistent piece completion
- failure to rediscover peers after removing and re-adding a torrent unless the whole process was restarted
- inability to reliably control TORRENT_STORAGE_DEFAULT_FILE_IO from the embedding application's own init or main because the environment was read too early
### 1. Classic file-storage optimization
The classic file I/O path was optimized to reduce avoidable filesystem churn and improve throughput under active download workloads.

Changes include:

- caching writable file handles in classicFileIo so chunk writes no longer repeatedly reopen the same files
- allowing storage reads used by hashing to reuse an already-open writer handle where possible
- adding file-level completion tracking so part-file promotion no longer needs repeated full-range piece scans
- fixing .part handling so UsePartFiles=false no longer touches part-file paths
- fixing false io.ErrShortWrite behavior on the classic read path during hashing
- expanding storage tests around rename behavior, borrowed read handles, fixed-length reads, and no-part-file flows
### 2. Buffered piece-completion checkpoints
Persistent piece completion was reworked to avoid doing synchronous durability work for every completed piece.

Changes include:

- wrapping Bolt-backed piece completion with an in-memory buffered overlay
- deferring complete=true persistence until checkpoint time while keeping complete=false persistence immediate
- adding PieceCompletionChange , PieceCompletionBatchSetter , and PieceCompletionCheckpointer interfaces
- adding SetBatch support for Bolt-backed completion persistence
- introducing a torrentCheckpointBuffer to batch file flushes and completion checkpoints
- checkpointing on a 1s / 128 MiB threshold instead of flushing every completed piece
- flushing buffered checkpoint state on torrent close
- protecting checkpoint queue bookkeeping with mutex-based snapshot/restore handling to avoid concurrent map iteration/write failures
- adding regression coverage for deferred persistence, close-time flush, and concurrent queueing
This reduces write-path overhead while preserving correctness and restart durability semantics.

### 3. Tracker re-add recovery fix
Tracker announce state is now properly reset when a torrent is removed and then re-added with the same tracker key.

Changes include:

- resetting stale announce lifecycle state on re-add
- forcing the next eligible tracker announce to send a fresh Started event
- adding regression coverage for the forced Started path
This fixes a case where a re-added torrent could fail to rediscover peers until the client process was restarted.

### 4. Lazy resolution of default file I/O from environment
Default file I/O backend selection was moved out of package-init-time logic and into runtime instantiation.

Changes include:

- resolving TORRENT_STORAGE_DEFAULT_FILE_IO when the storage backend is actually created
- keeping the existing defaultFileIo() entry point, but making it lazy
- removing the built-in Windows-specific fallback to classicFileIo when the env var is unset
- making the implicit default consistent across platforms when no override is provided
- adding tests to verify no-env default behavior and lazy environment lookup
This makes the library friendlier for embedding applications that want to set storage mode from their own startup flow.

## Impact
This PR is intended to improve:

- storage throughput on active downloads
- stability of persistent piece completion under heavy workloads
- correctness when re-adding torrents
- configurability for embedded consumers


storage/file-io-classic.go
```diff
@@ -1,48 +1,157 @@
...
 	*os.File
 }
 
 func (c classicFileReader) writeToN(w io.Writer, n int64) (written int64, err error) {
-	lw := limitWriter{
-		rem: n,
-		w:   w,
-	}
-	return c.File.WriteTo(&lw)
+	return io.CopyN(w, c.File, n)
```
anacrolix: Do you know if this avoids the unnecessary buffering? Was this a nonspecific change?

client-tracker-announcer.go
```diff
@@ -456,6 +452,17 @@ func (me *regularTrackerAnnounceDispatcher) addKey(key torrentTrackerAnnouncerKe
 		g.MapMustAssignNew(me.announceStates, key, g.PtrTo(announceState{}))
 	}
 	t.regularTrackerAnnounceState[key] = g.MapMustGet(me.announceStates, key)
+	if me.announceData.ContainsKey(key) {
+		me.resetAnnounceStateForReadd(key)
```
anacrolix, line 456: Readd?

client-tracker-announcer.go
```diff
@@ -859,6 +889,9 @@ type announceState struct {
 	lastAttemptCompleted time.Time
 	// Has ever sent completed event. Should only be sent once.
 	sentCompleted bool
+	// Force the next eligible announce to be a fresh Started, used when a torrent is re-added while
+	// stale tracker lifecycle state for the same key still exists.
+	forceStarted bool
```
anacrolix: I wonder if we can just distinguish what the last successful or attempted announce event was rather than special casing started?

anacrolix (commented):
This is a fantastic PR. Only a few questions to address.

anacrolix:
What's the DeMorgan's Law failure in the lint? Could you fix that?

storage/file-io-classic.go
Aodi: I originally thought that using io.CopyN would leverage zero-copy to improve performance. After checking the source code, I found it simply calls WriteTo in this scenario, so I reverted the relevant changes.

client-tracker-announcer.go
Aodi: > Readd?

It came to me that: remove and readd a torrent, AnnounceState may be not clear, so it's a check and reset.

client-tracker-announcer.go
Aodi: > I wonder if we can just distinguish what the last successful or attempted announce event was rather than special casing started?

Yes you are right, I think lastOk.Completed is ok for logic, so I remove the forceStarted code.
I will have a more deep download readd test on next monday or later.
I commit the code first, but It's best to wait my deep test.

Aodi:
> What's the DeMorgan's Law failure in the lint? Could you fix that?

Sure, I fixed it.

client-tracker-announcer.go
anacrolix: I mean do you intend "Read" or have I missed something

anacrolix:
Ok I'm happy with your attention to detail here, so let me know when you want to merge.

I'll defer to you on how to handle the last announce state, I see you removed forcestarted but you might want something to replace it.

Aodi:
> Ok I'm happy with your attention to detail here, so let me know when you want to merge.
> 
> I'll defer to you on how to handle the last announce state, I see you removed forcestarted but you might want something to replace it.

@anacrolix  I'm late to reply, sorry for that, what a busy week to me.

About `forceStarted`:  I analyzed the existing code paths and forceStarted was redundant because resetAnnounceStateForReadd always zeroes lastOk. Completed at the same time, so the `lastOk.Completed.IsZero()` check in nextAnnounceEvent already ensures a fresh Started event. I don't think there's a case where we need to force a Started event without also resetting lastOk , so I'd keep the change as-is. 

I tested new code, re-add, re-announce, everything is ok!

Go ahead and merge when you're ready.

anacrolix:
I shuffled things around on `master` due to some issues. You might want to rebase if you could.

Aodi:
> I shuffled things around on `master` due to some issues. You might want to rebase if you could.

Rebase work is done.

client-tracker-announcer.go
anacrolix: Could you clarify readd vs read?

anacrolix:
Thanks. Running tests. 

In future, smaller PRs are preferable, and stuff I can cherry pick (merging makes it hard).

anacrolix:
Haven't had time to look at the test failure, could you?

Aodi:
> Thanks. Running tests.
> 
> In future, smaller PRs are preferable, and stuff I can cherry pick (merging makes it hard).

I agree with you, I'm sorry for this pr, It's lots of code. I will notice it in future.

Aodi:
> Haven't had time to look at the test failure, could you?

The failed test, is because of revert code for `writeToN` in `classicFileReader`.

 c.File.WriteTo(&lw) returned `io.ErrShortWrite`.
 
 I think the io.ErrShortWrite returned by limitWriter is not an actual error, it's just how the limit signals that it has consumed its quota.
 
 I just remove the test now, it's optional test for this code, but I prefer to change code like following this:
 
```go
func (c classicFileReader) writeToN(w io.Writer, n int64) (written int64, err error) {
	lw := limitWriter{
		rem: n,
		w:   w,
	}
	written, err = c.File.WriteTo(&lw)
	if errors.Is(err, io.ErrShortWrite) {
		err = nil
	}
	return
}
```

simply treats io.ErrShortWrite from limitWriter as success.

If you agree the change, I will commit the code like this.

anacrolix:
I think the reason I didn't do this already was the contract for Write says an error must be returned if all bytes aren't written. Thanks for checking. I'll have a quick look and merge if it's ok.

Aodi:
The go test failed on ubuntu. Need I do something?

anacrolix:
@copilot resolve the merge conflicts in this pull request

anacrolix:
@copilot resolve the merge conflicts in this pull request

anacrolix:
This got stuck again, I think other PRs have nibbled some of the fixes. Could you break it up into smaller more digestable PRs?

Aodi:
> This got stuck again, I think other PRs have nibbled some of the fixes. Could you break it up into smaller more digestable PRs?

Sorry this got stuck for so long. I should have split it earlier.

That makes sense. I agree the original PR became too large and hard to keep current.

I've started splitting it into smaller PRs based on independent pieces of behavior. The first one is here:

- #1081

This one only covers the tracker re-add announce-state reset. I rebased it on current master, kept it to a single focused commit, and verified it with:

- `go test -timeout=5m ./...`

I'll continue with the storage changes as separate smaller PRs, keeping dependent changes ordered so each PR stays reviewable and cherry-pickable.

There will be new pr after #1081 end

Aodi:
I split code to small pr with AI's help, and I checked before pr created.

Aodi:
https://github.com/anacrolix/torrent/pull/1086
https://github.com/anacrolix/torrent/pull/1087
https://github.com/anacrolix/torrent/pull/1088

Here are New PRs. These three PRs have no overlapping files.

anacrolix:
I didn't investigate too much, but I assume the new PRs cover everything that was bundled in this PR, so I'll close this. Let me know if we missed something @Aodi. Thanks very much.

PR #1052, opened 2026-05-17
Complete BEP 6 Fast Extension: send Allowed Fast Set + record received

### LINKED ISSUE #235: Finish implementing BEP 6 Fast Extension
http://www.bittorrent.org/beps/bep_0006.html

guangtoutong:
Closes #235.

Completes BEP 6 Fast Extension by adding the two missing pieces:

### 1. Bug fix: received `AllowedFast` was a no-op

Before this change, `case pp.AllowedFast` in `peerconn.mainReadLoop` logged the message but **never wrote to `peerAllowedFast`**. Every downstream check (`requesting.go`, `shouldRequestWithoutBias`, `peer.go:peerAllowedFast.Iterate`, `torrent.go:1548`) read an empty bitmap, so honoring a peer's allowed-fast set was effectively disabled.

### 2. New: send our Allowed Fast Set

When `fastEnabled()` and torrent metadata are available, we now compute the BEP 6 allowed-fast set for the remote peer and send one `AllowedFast` message per piece after the Have-all/Have-none/Bitfield decision in `sendInitialMessages`. Peers can request these pieces while choked, removing the cold-start penalty.

### Algorithm

`internal/fast/fast.go` — SHA1-based pseudorandom indices seeded with the /24-masked peer IP and v1 info hash. Pure function so peers can independently re-derive and verify. Returns nil for IPv6 (BEP 6 defines the algorithm for IPv4 only). Reference: [cenkalti/rain/internal/fast](https://github.com/cenkalti/rain/blob/master/internal/fast/fast.go) (as suggested in #235).

### Tests

`internal/fast/fast_test.go`:

- **BEP 6 canonical example**: ip=80.4.4.200, infohash=0xAA*20, numPieces=1313, k=7 ⇒ `[1059, 431, 808, 1217, 287, 376, 1188]` ✅
- Determinism (peers must reproduce the same set)
- /24 subnet stability (same subnet ⇒ same set, different subnet ⇒ different)
- IPv6 returns nil
- Uniqueness in collision-prone configurations (k = numPieces)
- Edge cases: k=0, numPieces=0

All existing tests pass. `storage/possum` fails locally on Windows due to a missing native `libpossum` library — unrelated to this change.

### Files

- `internal/fast/fast.go` — new package
- `internal/fast/fast_test.go` — new tests
- `peerconn.go` — fix receive path, add `sendAllowedFastSet`
- `client.go` — invoke `sendAllowedFastSet` after Have-all/Have-none/Bitfield

cc @cenkalti (parts of the algorithm are based on rain's implementation).

### DIFF
--- a/client.go
+++ b/client.go
@@ -1338,6 +1338,12 @@ func (pc *PeerConn) sendInitialMessages() {
 		}
 		pc.postBitfield()
 	}()
+	// BEP 6: announce the Allowed Fast Set so the peer can request these pieces
+	// while still choked. Only meaningful when both sides support Fast Extension
+	// and we know how many pieces the torrent has.
+	if pc.fastEnabled() && t.haveInfo() {
+		pc.sendAllowedFastSet()
+	}
 	if pc.PeerExtensionBytes.SupportsDHT() && cl.config.Extensions.SupportsDHT() && cl.haveDhtServer() {
 		pc.write(pp.Message{
 			Type: pp.Port,
--- a/internal/fast/fast.go
+++ b/internal/fast/fast.go
@@ -1,0 +1,71 @@
+// Package fast implements the Allowed Fast Set generation algorithm from
+// BEP 6 (Fast Extension).
+//
+// http://www.bittorrent.org/beps/bep_0006.html
+//
+// The algorithm is a deterministic function of the peer's IPv4 (masked to /24)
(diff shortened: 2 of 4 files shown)


anacrolix:
Can you describe the Windows test failure? It should be possible to clone submodules and install the right toolchain and have those pass too. Open an issue if not.

PR #1053, opened 2026-05-17
Implement BEP 27 (private torrents)

guangtoutong:
Honors the \`private=1\` flag in torrent metainfo per [BEP 27](http://www.bittorrent.org/beps/bep_0027.html). Previously \`metainfo.Info.Private\` was parsed but never checked at runtime, so private trackers' workflow was silently leaking peers via DHT, PEX, and LSD.

### What changed

Adds a single helper \`Torrent.isPrivate()\` and consults it at every existing decentralized-discovery gate:

| Gate | File | Behavior for private |
|------|------|---------------------|
| DHT announce loop | \`torrent.go:dhtAnnouncer\` | Skipped per iteration |
| PEX startup | \`pexconn.go:pexConnState.Init\` | PEX not enabled |
| LPD outbound announce | \`client.go:TorrentInfohashesAndPort\` | Excluded from list |
| LPD inbound peer add | \`client.go:OnLPDAnnouncement\` | Both targeted and fallback paths skip |
| Post-add LPD kick | \`client.go:AddTorrentOpt\` | \`lpdPeers\`/\`lpdForce\` skipped |

### Magnet edge case

\`isPrivate()\` returns false when info has not yet been loaded (e.g. a magnet still fetching metadata). The downstream loops re-check each iteration, so the moment metadata arrives the flag takes effect. This is the only safe behaviour — we physically cannot honour a private flag we don't yet have. In practice private trackers distribute \`.torrent\` files (not magnets) because of this race.

### Tests

\`bep27_private_test.go\` covers the four meaningful states of the helper: no info, \`Private==nil\`, \`Private==&false\`, \`Private==&true\`.

Per-gate behaviour is exercised by the existing integration suite (\`./test/...\`), which still passes on every public-torrent path. \`storage/possum\` fails locally on Windows due to missing native \`libpossum\` — unrelated to this change.

### Spec reference

> If true, the client MUST publish its presence to get other peers ONLY via the trackers explicitly described in the metainfo file. If false or is not present, the client may obtain peer from other means, i.e. PEX peer exchange, DHT.
>
> — BEP 27

### DIFF
--- a/bep27_private_test.go
+++ b/bep27_private_test.go
@@ -1,0 +1,55 @@
+// BEP 27 private-torrent semantics.
+// http://www.bittorrent.org/beps/bep_0027.html
+package torrent
+
+import (
+	"testing"
+
+	"github.com/anacrolix/torrent/metainfo"
+)
+
+// makeTorrentWithPrivate is a focused helper for the BEP 27 tests below.
+// It builds a *Torrent shell carrying just the info needed by isPrivate(),
+// without touching the network or storage. The Torrent is not registered
+// with a Client and must not be used for any other operations.
+func makeTorrentWithPrivate(p *bool) *Torrent {
+	return &Torrent{
+		info: &metainfo.Info{Private: p},
+	}
+}
+
+func TestIsPrivate_NoInfo(t *testing.T) {
+	// Before metadata is loaded (e.g. magnet still fetching), isPrivate
+	// returns false. The downstream loops re-check on every iteration,
+	// so once info arrives the privacy flag takes effect immediately.
+	tor := &Torrent{}
+	if tor.isPrivate() {
+		t.Fatal("torrent with no info should not report private")
+	}
+}
+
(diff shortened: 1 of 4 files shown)


anacrolix:
This is very good, thanks.

What do you think about adding an extra check that peers aren't added or used after torrent privacy is determined? I see your point about needing to treat a torrent as public up until you know it isn't, to discover the metainfo, which means we can't default to private until otherwise (the safest way to handle this), but we can also assume the stakes around privacy are high enough that we want to abort or error if accidental usage occurs that would violate the privacy too (i.e. make it crash or error).

anacrolix:
So if you follow my drift, put private checks in a few key places where we should not be (pending peers from non-tracker sources, DHT?) and force a panic or error. None of this blocks this PR which is correct and an improvement regardless, so you could do it in a subsequent PR if it interests you.

PR #1055, opened 2026-05-24
tracker/http: handle malformed non-compact peer list entries

Sahil-4555:
### Issue
The non-compact peers list parser in `Peers.UnmarshalBencode` does a bare type assertion on each list element - `i.(map[string]interface{})` - without checking if the element is actually a dictionary. Inside `FromDictInterface`, the `ip` and `port` fields are also pulled out with unchecked assertions that panic on missing or wrong-typed values. So if a tracker sends back something like `peers: [42]` instead of the expected list of dictionaries, the client panics and crashes. Same thing happens if a dictionary is missing `ip` or `port`, or if those fields have the wrong types. Since tracker URLs come from untrusted torrent files and magnet links, and announces run automatically in the background, a single bad tracker response is enough to take down the whole process.

### Fix
Both layers now use safe comma-ok type assertions instead of bare ones. In `UnmarshalBencode`, each list element is checked before treating it as a peer dict. If it's not a dictionary, we just skip it - this way a response with some valid and some garbage peers still returns the good ones instead of failing entirely.

`FromDictInterface` now returns an error instead of panicking. It validates that `ip` is a string and actually parseable by `net.ParseIP` (so we don't end up with nil IPs in the system), and that `port` is an integer. `peer id` stays optional as per BEP 3. The errors are package-level sentinels using `errors.New` - no `fmt.Errorf` allocations per call, which matters when a hostile tracker sends many bad entries.

One subtle thing we caught while writing this: the original loop used `err = p.FromDictInterface(...)` which writes to the named return. If the last entry in the list was invalid, that error would leak out through the bare `return` even though we collected valid peers above it. Fixed by using `err :=` to keep it local, with a comment so nobody changes it back by mistake.

### Tests
Unit tests use a table-driven `TestUnmarshalPeersMalformed` with six subcases:

- **NonDictInList**: integer in peers list, the original crash trigger
- **MixedValidAndInvalid**: valid peers mixed with garbage, checks that good ones are still kept
- **MissingFields**: dict with only `peer id`, no `ip` or `port`
- **WrongFieldTypes**: `ip` as integer, `port` as string
- **EmptyList**: empty list, normal tracker behaviour
- **InvalidIP**: `ip` is a string but not a valid address, checks the `net.ParseIP` guard

### DIFF
--- a/tracker/http/http_test.go
+++ b/tracker/http/http_test.go
@@ -73,3 +73,59 @@ func TestSetAnnounceInfohashParamWithSpaces(t *testing.T) {
 		qt.StringContains(someUrl.String(),
 			"info_hash=%2Bv%0A%A1x%93%200%C8G%DC%DF%8E%AE%BFV%0A%1B%D1l"))
 }
+
+// These cases cover various forms of malformed non-compact peer lists that a
+// tracker might return. Before the fix, several of these caused panics due to
+// bare type assertions on untrusted data.
+func TestUnmarshalPeersMalformed(t *testing.T) {
+	cases := []struct {
+		name      string
+		input     string
+		wantCount int
+	}{
+		{
+			// Integer in the peers list instead of a peer dictionary.
+			name:      "NonDictInList",
+			input:     "d8:intervali1800e5:peersli42eee",
+			wantCount: 0,
+		},
+		{
+			// Two valid peer dicts with an invalid integer between them.
+			// Both valid peers should still be collected.
+			name:      "MixedValidAndInvalid",
+			input:     "d5:peersld2:ip7:1.2.3.44:porti6881eei99ed2:ip8:5.6.7.894:porti6882eeee",
+			wantCount: 2,
+		},
+		{
+			// Peer dict has "peer id" but is missing the required "ip" and "port".
+			name:      "MissingFields",
+			input:     "d5:peersld7:peer id4:testeee",
+			wantCount: 0,
+		},
+		{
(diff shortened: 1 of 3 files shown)


anacrolix:
Not blockers, but rather than table-driven tests, could/would you do function sub tests passing the the case for each call site? It's typically overlooked but much better to debug.

Also it's probably not worth creating new error types for non-actionable errors (like who would worry if it's no IP vs no port for example.

Keen to know your thoughts on those.

Sahil-4555:
> Not blockers, but rather than table-driven tests, could/would you do function sub tests passing the the case for each call site? It's typically overlooked but much better to debug.
> 
> Also it's probably not worth creating new error types for non-actionable errors (like who would worry if it's no IP vs no port for example.
> 
> Keen to know your thoughts on those.

honestly table-driven felt like the natural fit here since all cases share the same two-line assertion logic (unmarshal + check count), no per-case branching to step through. But yeah looking at the rest of http_test.go, individual named tests is already the convention in this file, so makes sense to restructure to match that. Coverage stays the same, just the shape changes.
 
you're right about errors which i add. Nobody actually inspects which error it is - the caller just checks err != nil and skips. Having two separate sentinels for a distinction that no consumer differentiates is unnecessary granularity. Will collapse them into a single errMalformedPeerDict. Still avoids the per-call fmt.Errorf allocation, just without the false precision.

Let me know if you'd like these changes, I can put up a follow-up PR for it.?

anacrolix:
It's barely worth it, just wanted to flag it. Thanks again :)

PR #1056, opened 2026-06-02
fix BEP 52 metainfo handling

Sahil-4555:
## Issue

This one came up while going through the BEP 52 metainfo handling paths. `FileTree.PiecesRootAsByteArray` copies the bencoded `pieces root` into a `[32]byte` and straight up panics if the length isn't exactly 32. The thing is, this gets called from `upvertedFilesInner`, which is the file tree iterator used by pretty much everything downstream - `FileSegmentsIndex()`, `initFiles()`, `cacheLength()`, `Piece.Length()`, all of them go through it.

Now, `validateInfo` is supposed to be the gate that catches bad metainfo before `setInfo` starts using it. But it doesn't actually check anything about the v2 file tree entries. So what happens is `validateInfo` passes fine, then `setInfo` calls `info.FileSegmentsIndex()`, the iterator walks into the file tree, hits `PiecesRootAsByteArray`, and the whole thing panics on the malformed root.

The annoying part is this is reachable through every public ingestion path - `AddTorrent`, `AddTorrentFromFile`, `AddTorrentSpec`, `SetInfoBytes`, and even through peer metadata exchange via `setInfoBytesLocked`. The panic fires before any piece transfer happens, right during the initial torrent setup itself.

For anyone running crawlers, gateways, or any long-running service that picks up untrusted torrents, one bad torrent file is enough to bring down the whole process. That's not great.

## Fix

Added a `FileTree.Validate()` method in `metainfo/file-tree.go`. It walks the v2 file tree and checks the BEP 52 structural requirements - non-empty files need exactly 32 bytes for the pieces root, and empty files shouldn't have one at all.

This gets called right at the top of `validateInfo()` in `misc.go`, before anything else touches the file tree. If the tree is malformed, `setInfo` returns a proper error and the torrent gets rejected cleanly - no storage gets opened, no files get initialised, no pieces get created.

The `panic(n)` in `PiecesRootAsByteArray` stays as it is. After the validation runs, it should never fire from untrusted input anymore. It's just a defensive invariant at that point, which is consistent with how `panicif` is used elsewhere in the codebase.

Kept the fix intentionally small. No signature changes to `PiecesRootAsByteArray`, no new allocations on the hot path, no changes to file iteration semantics. Valid BEP 52 torrents keep working exactly as before.

## Tests

Added coverage in the existing relevant test files:

- `TestFileTreeValidate` - table-driven test in `metainfo/info_test.go` for the new `Validate()` method. Covers valid files, empty files, short roots, long roots, missing roots on non-empty files, empty files with spurious roots, and nested directories with mixed good and bad entries.
- `TestValidateInfoShortV2Root` - checks that `validateInfo` rejects a v2 info dict where the pieces root is too short.
- `TestValidateInfoMissingV2Root` - checks that `validateInfo` catches a non-empty file with no pieces root at all.
- `TestValidateInfoValidV2` - makes sure valid v2 info still goes through fine, so we haven't broken the happy path.
- `TestAddTorrentSpecBadV2Root` - end-to-end through the public `AddTorrentSpec` API with crafted bencoded info bytes. Confirms the error comes back cleanly instead of panicking.

### DIFF
--- a/metainfo/file-tree.go
+++ b/metainfo/file-tree.go
@@ -1,6 +1,7 @@
 package metainfo
 
 import (
+	"fmt"
 	"iter"
 	"maps"
 	"slices"
@@ -156,4 +157,25 @@ func (ft *FileTree) PiecesRootAsByteArray() (ret g.Option[infohash_v2.T]) {
 	}
 	ret.Ok = true
 	return
+}
+
+// Validate checks structural invariants required by BEP 52. Non-empty leaf files must have a
+// pieces root of exactly 32 bytes; empty files must not have one.
+func (ft *FileTree) Validate() error {
+	var err error
+	ft.Walk(nil, func(path []string, node *FileTree) {
+		if err != nil {
+			return
+		}
+		if node.IsDir() {
+			return
+		}
+		n := len(node.File.PiecesRoot)
+		if node.File.Length > 0 && n != 32 {
+			err = fmt.Errorf("file %q: pieces root length %d, expected 32", path, n)
+		} else if node.File.Length == 0 && n != 0 {
+			err = fmt.Errorf("file %q: empty file has pieces root of length %d", path, n)
+		}
+	})
+	return err
 }
--- a/metainfo/info_test.go
+++ b/metainfo/info_test.go
@@ -1,6 +1,7 @@
 package metainfo
 
 import (
+	"strings"
 	"testing"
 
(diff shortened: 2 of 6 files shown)


anacrolix:
I've been a nitwit and rewritten master which gave you an extra commit that doesn't belong here.

anacrolix:
I rebased without squash so we can remove the extra commit if needed. Thank you!

PR #1057, opened 2026-06-11
versionToChar: return byte

milahu:
rune can be multiple bytes
but here we need exactly one byte

continue #1026


anacrolix:
Awesome

PR #1063, opened 2026-07-10
feat: add ClientConfig.ListenPacket

hanselime:
Adds an optional `ListenPacket func(network, addr string) (net.PacketConn, error)` to `ClientConfig`. When set, the uTP socket and the DHT server that shares it, is built on the returned `net.PacketConn` instead of one bound directly to the host network, so uTP and DHT traffic can be tunneled (e.g. through a userspace WireGuard/netstack conn).

the cgo (go-libutp) path depends on `NewSocketFromPacketConn` from https://github.com/anacrolix/go-libutp/pull/34, so go.mod needs that release before merge.


anacrolix:
go-libutp is updated on torrent/master now. The CI should pass now. When it does we can merge.

PR #1069, opened 2026-07-28
Arm peer message writer wakeup before refilling its buffer

softganz88:
## What changed

- arm `writeCond` before `fillWriteBuffer` evaluates whether more peer messages are available
- add a deterministic regression that forces a broadcast during the first empty refill and requires a second refill

## Why

`BroadcastCond` does not retain a broadcast when no `Signaled` channel is armed. The peer message writer previously called `fillWriteBuffer` before arming `writeCond`, while the refill observes peer state outside `cn.mu`. A state transition could therefore broadcast during refill, leave the write buffer empty, and be lost before the writer began waiting. With no later transition, the writer remained asleep.

Arming the condition first retains a broadcast from that refill window without changing the existing closed-channel or keepalive selection.

## Validation

- `go test -run '^TestPeerConnMsgWriterArmsWakeupBeforeRefill$' -count=100 .`
- `cargo build` in the pinned `storage/possum/lib` submodule, then `go test ./...`
- `gofmt` and `git diff --check`

The regression uses channel synchronization and contains no timing sleep for the target interleaving. It fails on commit `8c67afb7185d42ea94dcced246596d79d35feb64` and passes after this change.

### DIFF
--- a/peer-conn-msg-writer.go
+++ b/peer-conn-msg-writer.go
@@ -84,6 +84,10 @@ func (cn *peerConnMsgWriter) run(keepAliveTimeout time.Duration) {
 		if cn.closed.IsSet() {
 			return
 		}
+		// Arm the condition before evaluating the predicate. BroadcastCond does
+		// not retain a broadcast when no channel is armed, and fillWriteBuffer
+		// runs without cn.mu while it observes peer state under another lock.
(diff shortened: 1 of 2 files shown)


anacrolix:
I went with #1078 instead. Thanks!

PR #1071, opened 2026-08-03
fix(torrent): rearm the msg writer keepalive timer before parking

### LINKED ISSUE #1070: Peer request-update wakeup can be lost by the message writer, wedging the connection
Three pieces interact on current master:

1. `peer-conn-msg-writer.go` `run` subscribes after checking: `fillWriteBuffer()` (consumes `needRequestUpdate`) runs before `writeCond.Signaled()`. `chansync.BroadcastCond.Broadcast` with no installed channel is a no-op, so a tickle landing in between is dropped.
2. `peerconn.go` `onNeedUpdateRequests` guards on `needRequestUpdate != ""`, so a dropped tickle is never retried.
3. `keepAliveTimer` is only `Reset` after a successful write; a bare fire on an idle connection leaves it dead, so there is no periodic backstop.

Most reachable when a peer attaches during initial piece verification: everything is ignored for requests mid-hash, so the single tickle at settle time is that connection's only wakeup. Result: a download wedges at zero bytes with a live, unchoked seeder.

aslafy-z:
Follow-up to #1078, [as suggested](https://github.com/anacrolix/torrent/pull/1071#issuecomment-5384689612). The wakeup-channel half of this PR landed there, so this is rebased onto master and reduced to the keepalive timer fix.

## Change

`keepAliveTimer` was only `Reset` after a successful write. One bare fire on an idle connection (nothing buffered, `keepAlive()` false) left the timer dead, so the periodic keepalive backstop was gone for the rest of the connection's life.

It is now armed before each park, with the time remaining until the keepalive deadline, falling back to a full interval once that deadline has passed. Arming per park also keeps the deadline accurate, where the old placement restarted the interval on every write.

## Tests

`TestMsgWriterKeepAliveTimerRearmed` lets the timer fire once while the connection is not yet useful, then makes it useful. No tickle is involved, so the keepalive can only come from a re-armed timer. It fails on master with `keepalive never written: timer was not rearmed after firing`, and passes under `-race` with the fix.

Most of the diff is the two helpers the test shares with the file, `msgWriterSink` and `startMsgWriter`.

## External validation

#1078 on its own was not enough downstream. The re-arm here was needed on top of it.

DataHearth/streamline#4 pins master plus this commit. Their bittorrent suite passes, 34/34, including a 30-download-cycle spec that reliably wedged under CPU contention on v1.61.

### DIFF
--- a/peer-conn-msg-writer.go
+++ b/peer-conn-msg-writer.go
@@ -84,6 +84,8 @@ func (cn *peerConnMsgWriter) run(keepAliveTimeout time.Duration) {
 		if cn.closed.IsSet() {
 			return
 		}
+		// Subscribe before checking for work, so a Broadcast landing while filling isn't dropped.
+		writeCond := cn.writeCond.Signaled()
 		cn.fillWriteBuffer()
 		keepAlive := cn.keepAlive()
 		cn.mu.Lock()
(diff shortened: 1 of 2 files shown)


anacrolix:
I accepted #1078 it was more immediately useful. Please feel free to submit the keep alive fix separately.

PR #1075, opened 2026-08-08
dialer: add SOCKS5 proxy support for peer connections

### LINKED ISSUE #160: SOCKS5 support for peer connections
Hi, I am still browsing through this package, I don't see a way to configure proxy. Can I do something with golang.org/x/net/proxy , I am interested in socks5 proxy.

Thanks

Pastalikek65:
## Issue

Closes #160 — SOCKS5 support for peer connections.

Currently only `ClientConfig.HTTPProxy` exists, which applies to HTTP tracker requests. Peer connections are dialed directly via `Client.dialers`. This PR adds a SOCKS5 dialer that can be registered with `Client.AddDialer`, so outgoing peer connections are routed through a SOCKS5 proxy.

## Changes

- `dialer/socks5.go`: new `Socks5` dialer (RFC 1928/1929) built on `golang.org/x/net/proxy`, which is already a dependency. Supports optional username/password auth and a custom forward dialer for connecting to the proxy itself.
- `dialer/socks5_test.go`: a minimal in-process SOCKS5 server exercising the full CONNECT handshake and an echo roundtrip, plus a connection-error case and a `DialerNetwork` sanity check.
- `README.md`: usage snippet showing `cl.AddDialer(dialer.NewSocks5("localhost:1080", nil))`.

## Verification

- `go test ./dialer/` — all pass (3 new tests).
- `gofmt`, `go vet ./dialer/` clean.

Note: listening for incoming connections is unaffected; only outgoing peer dials are proxied, consistent with how `AddDialer` works today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,10 +43,12 @@ There are some small [examples](https://pkg.go.dev/github.com/anacrolix/torrent#
 
 ### Routing peer connections through a SOCKS5 proxy
 
-To dial peers through a SOCKS5 proxy, add the [dialer.Socks5](https://pkg.go.dev/github.com/anacrolix/torrent/dialer#NewSocks5) dialer to the client:
+To dial peers through a SOCKS5 proxy, add the [dialer.Socks5](https://pkg.go.dev/github.com/anacrolix/torrent/dialer#NewSocks5) dialer to the client. Start from the default config and set `DialForPeerConns` to false so the SOCKS5 dialer is the only one used for outgoing peer connections:
 
 ```go
-cl, _ := torrent.NewClient(&torrent.ClientConfig{})
+cfg := torrent.NewDefaultClientConfig()
+cfg.DialForPeerConns = false
+cl, _ := torrent.NewClient(cfg)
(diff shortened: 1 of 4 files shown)


anacrolix:
Can you address the copilot comments?

Pastalikek65:
Both copilot points are addressed in cb70e80e.

1. It was right about the empty ClientConfig — with all defaults nil, init panics on `UploadRateLimiter.Burst()` in `setRateLimiterBursts` (config.go:298). I reproduced the crash before touching anything.

2. The README example now starts from `NewDefaultClientConfig` with `DialForPeerConns = false`, so the SOCKS5 dialer is the only dialer used for outgoing peer connections while incoming listeners stay as they are.

Also added `TestClientWithSocks5Dialer` to client_test.go, which builds the client exactly the way the example shows (config → NewClient → AddDialer → Close). That test, the dialer suite, and vet are green locally; full CI runs on the PR itself.

anacrolix:
Awesome, thanks

PR #1079, opened 2026-08-23
announce dispatcher: don't record announce keys that couldn't be created

### LINKED ISSUE #1068: panic: false in regularTrackerAnnounceDispatcher.updateTorrentInput — announce key recorded on Torrent but never registered
Adding a torrent that carries a v2 infohash reliably panics the process from the announce timer goroutine, a second or two after the add. Because it fires on `regularTrackerAnnounceDispatcher`'s own goroutine, an embedding application cannot recover from it.

```
panic: false

goroutine 262058 [running]:
github.com/anacrolix/missinggo/v2/panicif.False(...)
	missinggo/v2@v2.10.0/panicif/panicif.go:73
github.com/anacrolix/torrent.(*regularTrackerAnnounceDispatcher).updateTorrentInput(...)
	client-tracker-announcer.go:583
github.com/anacrolix/torrent.(*regularTrackerAnnounceDispatcher).step(...)
	client-tracker-announcer.go:429
github.com/anacrolix/torrent.(*regularTrackerAnnounceDispatcher).timerFunc(...)
	client-tracker-announcer.go:418
github.com/anacrolix/torrent/internal/mytimer.(*Timer).innerCallback(...)
	internal/mytimer/my-timer.go:69
created by time.goFunc
```

Seen at `v1.61.1-0.20260622020128-fd15081f8ed1`. `client-tracker-announcer.go` appears unchanged since 5d662f77 (2026-03-15), so this should still be present at HEAD. It is also in the `v1.61.0` tag.

### Mechanism

`updateTorrentInput` walks `t.regularTrackerAnnounceState` and asserts every key still has a row in `announceData`:

```go
res := me.announceData.Update(key, func(av nextAnnounceInput) nextAnnounceInput { ... })
panicif.False(res.Exists)
```

Nothing ever removes rows: there is no `delete`/`clear` on `announceData`, `announceStates` or `Torrent.regularTrackerAnnounceState` anywhere in the module, and rows are only created by `addKey`, which writes `announceStates[key]` and `announceData.Create(key, ...)` back to back. So a key on the Torrent with no row can only be a key that was **never registered**.

`addKey` declines exactly one way — when the short infohash has no entry in `torrentsByShortHash`:

```go
t := me.torrentFromShortInfohash(key.ShortInfohash)
if t == nil {
    // Crude, but the torrent was already dropped. We probably called AddTrackers late.
    return
}
```

but `Torrent.initRegularTrackerAnnounceState` records the key regardless, storing the nil `*announceState` the failed lookup returns.

### The route I reproduced

`Torrent.eachShortInfohash` yields the truncated **v2** infohash as well as the v1 one, and announcing starts a tracker per short infohash — but nothing registers the v2 short hash in `torrentsByShortHash` when both hashes are known at add time. So any torrent carrying a v2 infohash strands a key on the first announce pass, and the next timer tick panics.

Adding a hybrid torrent (or a magnet with both `xt` values) and waiting for the announce timer reproduces it deterministically.

### Suggested direction

Two small halves, which is what I have applied locally:

1. **Prevent** — have `addKey` report whether the key ended up registered, and have `initRegularTrackerAnnounceState` record nothing when it was declined (the announcer it would publish dereferences the nil `*announceState` anyway).
2. **Heal** — in `updateTorrentInput`, drop keys with no `announceData` row (and their `Torrent.trackerAnnouncers` sibling) with a logged warning, rather than asserting. Since nothing ever deletes rows, a missing row means the key was never registered, so there is no announce to lose.

Happy to open a PR if that direction looks right.

*Reported from HiveTorrent, a Wails desktop client. The panic was caught in production only after we added crash logging — in a `-H windowsgui` build the runtime's panic output goes to a stderr that does not exist, so the process simply vanished.*

Pastalikek65:
Fixes #1068.

## What

`initRegularTrackerAnnounceState` recorded the announce key on the torrent and deferred an input update unconditionally, even when `addKey` bailed out early. `addKey` returns early when `torrentFromShortInfohash(key.ShortInfohash)` is nil — which happens for the **v2** short infohash of a v2/hybrid torrent added without info bytes, because `torrentsByShortHash` only gains the v2 short hash later, in `hashInfoBytes`, when metadata arrives.

The deferred `updateTorrentInput` then iterates the recorded key and hits `panicif.False(res.Exists)` on the missing `announceData` entry — panicking the process from the dispatcher's timer goroutine, unrecoverable for embedding applications. The stack trace in the issue matches exactly.

## Fix

`addKey` now reports whether it registered the key, and `initRegularTrackerAnnounceState` only records the state and defers the update when it did. The `panicif` invariant in `updateTorrentInput` stays: every recorded key now genuinely has an `announceData` entry.

## How tested

New `TestV2InfohashUpdateTorrentInputNoPanic` (`client-tracker-announcer_test.go`): adds a hybrid (v1+v2 infohash) torrent with no info bytes plus a tracker, then runs an input update. On master this panics with `panic: false` at `client-tracker-announcer.go:583` via `step → timerFunc` — verified before applying the fix; with the fix it passes repeatedly. `go build ./...` clean.

Note: I could not run the full package test suite or `-race` in my environment (aarch64 host with 39-bit VMA — the race detector refuses with "unsupported VMA range" and large test binaries segfault intermittently, reproducible on unmodified master), so verification is via the focused tests above.

Follow-up worth considering: registering the v2 short hash in `torrentsByShortHash` at `AddTorrentOpt` time would make v2-key announces work even before metadata arrives, instead of being skipped until then.

### DIFF
--- a/client-tracker-announcer.go
+++ b/client-tracker-announcer.go
@@ -439,14 +439,16 @@ func (me *regularTrackerAnnounceDispatcher) step() mytimer.TimeValue {
 	return me.nextTimerDelay()
 }
 
-func (me *regularTrackerAnnounceDispatcher) addKey(key torrentTrackerAnnouncerKey) {
+func (me *regularTrackerAnnounceDispatcher) addKey(key torrentTrackerAnnouncerKey) bool {
 	if me.announceData.ContainsKey(key) {
-		return
+		return true
 	}
 	t := me.torrentFromShortInfohash(key.ShortInfohash)
 	if t == nil {
-		// Crude, but the torrent was already dropped. We probably called AddTrackers late.
-		return
+		// Crude, but the torrent was already dropped. We probably called AddTrackers late. The
+		// caller must not record the key in that case, or updateTorrentInput will panic on the
+		// missing announceData entry.
+		return false
 	}
 	g.MakeMapIfNil(&me.torrentForAnnounceRequests)
 	// This can be duplicated when there's multiple trackers for a short infohash. That's fine.
@@ -462,6 +464,7 @@ func (me *regularTrackerAnnounceDispatcher) addKey(key torrentTrackerAnnouncerKe
 		infohashActive:         g.OptionFromTuple(me.infohashAnnouncing.Get(key.ShortInfohash)).Value.count,
 	})
 	me.updateTimer()
+	return true
 }
 
 // Returns nil if the torrent was dropped.
--- a/client-tracker-announcer_test.go
+++ b/client-tracker-announcer_test.go
@@ -9,9 +9,38 @@ import (
 	"testing/synctest"
 	"time"
 
+	g "github.com/anacrolix/generics"
 	"github.com/anacrolix/missinggo/v2/panicif"
 	"github.com/go-quicktest/qt"
+
+	"github.com/anacrolix/torrent/metainfo"
+	infohash_v2 "github.com/anacrolix/torrent/types/infohash-v2"
 )
+
+// Adding a torrent with a v2 infohash but no info bytes used to panic in
+// updateTorrentInput: addKey can't find the torrent by its v2 short hash
+// (not yet in torrentsByShortHash), returns early, but
+// initRegularTrackerAnnounceState still records the key and defers an
+// update. The subsequent updateTorrentInput then hits a key missing from
+// announceData. See #1068.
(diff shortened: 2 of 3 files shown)


anacrolix:
I'm not sure how you would register the V2 hash if you don't know it yet, unless I've missed something

PR #1082, opened 2026-08-25
peerconn: reuse chunk pool buffers for peer request data

AskAlexSharov:
`readPeerRequestData` allocated `r.Length` bytes for every served peer request, so seeding allocated ~16KiB per chunk uploaded. It now takes the buffer from the `Torrent` chunk pool (already used by the wire decoder) when the request fits, and returns it once the message writer has copied the piece data into its write buffer. Requests larger than the chunk size still allocate.

Release points: read error, request no longer wanted, `deleteReadyPeerRequest`, `deleteAllPeerRequests`, and after `msg()` in `sendChunk`. `sendChunk` uses a new `takeReadyPeerRequest` (delete without release) so it can't free the buffer before the copy. `PeerConn.onClose` now calls `deleteAllPeerRequests` so buffers go back to the pool and `torrentChunkBuffersOutstanding` stays honest.

`BenchmarkServePeerRequests` (new, drives `onReadRequest` -> `servePeerRequest` -> `sendChunk`), n=8:

|          | before | after |
|----------|--------|-------|
| B/op     | 17227  | 852 (-95%) |
| sec/op   | 3.28µs | 1.84µs (-44%) |

`-race` is clean on the root package, `./test/...` and `./tests/...`; a temporary check showed `torrentChunkBuffersOutstanding == 0` after real seed/leech transfers.

Carries #1083 (test lock fix) so CI here is green; master is red without it. Drops out on rebase once #1083 merges.

### DIFF
--- a/client-tracker-announcer_test.go
+++ b/client-tracker-announcer_test.go
@@ -36,9 +36,9 @@ func TestV2InfohashUpdateTorrentInputNoPanic(t *testing.T) {
 		InfoHashV2: g.Some(v2),
 	})
 	qt.Assert(t, qt.IsTrue(new_))
-	torr.addTrackers([][]string{{"http://tracker.example.com:6969/announce"}})
 	cl.lock()
 	defer cl.unlock()
+	torr.addTrackers([][]string{{"http://tracker.example.com:6969/announce"}})
 	cl.regularTrackerAnnounceDispatcher.updateTorrentInput(torr)
 }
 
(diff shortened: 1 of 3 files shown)


anacrolix:
Interesting. I think I optimized the chunk allocation recently but it must have only been for the read path. I think the issue with using the pool on the seeding/write path is we can't predict the remote peer will request consistently sized chunks, or chunks that match our preferred read chunk size. I'll have to read more carefully.

I'll update the branch to catch fixes that went in already too.

PR #1083, opened 2026-08-25
client-tracker-announcer: hold the client lock in TestV2InfohashUpdateTorrentInputNoPanic

AskAlexSharov:
`master` is red: `TestV2InfohashUpdateTorrentInputNoPanic` calls `torr.addTrackers` without the client lock. `addTrackers` mutates client-wide dispatcher state (`pendingTorrentInputUpdates` via `updateWantPeersEvent` -> `deferUpdateRegularTrackerAnnouncing`), so it races the dispatcher timer goroutine reading the same map in `step()`. `Torrent.AddTrackers` takes the lock for exactly this reason. Fix is to move the existing `cl.lock()` above the call.

```
go test -race -count 200 -run TestV2InfohashUpdateTorrentInputNoPanic .
```

9 `WARNING: DATA RACE` before, 0 after. CI hit it on `test (stable, ubuntu-latest)` and `test (oldstable, ubuntu-latest)`.

The `indexed.table.Poison` panic in `updateTorrentInput` on the master run at 7e8d409 is likely the same unsynchronised access seen from the timer side, though I didn't reproduce that one.

### DIFF
--- a/bep14_test.go
+++ b/bep14_test.go
@@ -79,13 +79,13 @@ func TestDiscovery(t *testing.T) {
 	qt.Assert(t, qt.IsNil(err))
 	t.Cleanup(func() { client1.Close() })
 	setupTestLPD(client1)
-	testutil.ExportStatusWriter(client1, "1", t)
+	defer testutil.ExportStatusWriter(client1, "1", t)()
 
 	client2, err := NewClient(config)
 	qt.Assert(t, qt.IsNil(err))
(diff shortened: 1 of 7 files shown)


anacrolix:
I think possibly someone else has picked up this fix already. Updating branch to see.

anacrolix:
Looks like it was already included. Thanks :)

PR #1085, opened 2026-08-25
request-strategy: measure allocations with AllocsPerRun

AskAlexSharov:
`TestRequestStrategyPieceDoesntAlloc` bracketed one call with `runtime.ReadMemStats` and asserted `HeapAlloc` was unchanged. `HeapAlloc` is process-wide, so any concurrent goroutine allocating between the two reads fails it. It went red on `test (oldstable, macos-latest)`:

```
got:  uint64(6761792)
want: uint64(6761984)
```

`HeapAlloc` went *down*, which no allocation in the call under test can cause. `testing.AllocsPerRun` measures only the function passed to it.

Still catches a regression: adding an allocation inside the closure fails the test. Clean at `-race -count 200`.

### DIFF
--- a/request-strategy-impls_test.go
+++ b/request-strategy-impls_test.go
@@ -2,7 +2,6 @@ package torrent
 
 import (
 	"context"
-	"runtime"
 	"testing"
 
 	g "github.com/anacrolix/generics"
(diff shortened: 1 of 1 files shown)


anacrolix:
Legend, thanks.

PR #1086, opened 2026-08-25
storage: skip part-file promotion when disabled

Aodi:
`MarkComplete` calls `promotePartFile` after every file becomes complete. When part files are disabled, `promotePartFile` still flushed the synthetic `.part` path before checking the option, causing unnecessary filesystem access and warnings for a path that should not be used.

Return before touching the `.part` path when `UsePartFiles` is disabled.

This change adds a regression test that verifies neither flush nor rename is attempted in that case.

Tests:

- `go test -timeout=5m ./...`

### DIFF
--- a/storage/file-piece.go
+++ b/storage/file-piece.go
@@ -238,14 +238,14 @@ func (me *filePieceImpl) MarkNotComplete() (err error) {
 }
 
 func (me *filePieceImpl) promotePartFile(f file) (err error) {
-	// Flush file on completion, even if we don't promote it.
+	if !me.partFiles() {
+		return nil
+	}
+	// Flush the part file before promotion so its contents are durable.
 	err = me.t.io.flush(f.partFilePath(), 0, f.length())
 	if err != nil {
 		me.logger().Warn("error flushing file before promotion", "file", f.partFilePath(), "err", err)
 		err = nil
-	}
-	if !me.partFiles() {
-		return nil
 	}
 	f.mu.Lock()
 	defer f.mu.Unlock()
--- a/storage/file-piece_partfiles_test.go
+++ b/storage/file-piece_partfiles_test.go
@@ -1,0 +1,75 @@
+package storage
+
+import (
+	"log/slog"
+	"testing"
+
+	"github.com/go-quicktest/qt"
+
+	g "github.com/anacrolix/generics"
+	"github.com/anacrolix/torrent/metainfo"
+)
+
+type trackingFileIo struct {
+	flushed []string
+	renamed bool
+}
+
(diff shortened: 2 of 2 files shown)


anacrolix:
Do you think perhaps it should have been flushing whatever the proper write file is, which when part files are disabled is the correctly named file? I think that was the original intent. I do see the issue with flushing part files if they're not in use tho. I will merge anyway under that assumption but I would like to know your opinion about flushing otherwise.

anacrolix:
Merging, there's another PR floating around that fixes the lint.

Aodi:
> Do you think perhaps it should have been flushing whatever the proper write file is, which when part files are disabled is the correctly named file? I think that was the original intent. I do see the issue with flushing part files if they're not in use tho. I will merge anyway under that assumption but I would like to know your opinion about flushing otherwise.

Yes, I think you're right. I interpreted `promotePartFile` too narrowly as only handling part-file promotion.

The invalid `.part` access was still a bug, but instead of skipping the flush entirely, we should flush the actual write path. `pathForWrite` already resolves to the `.part` file when part files are enabled and the correctly named file when they are disabled. Only the rename should be conditional on `partFiles()`.

Something like this:

```go
func (me *filePieceImpl) promotePartFile(f file) (err error) {
	writePath := me.t.pathForWrite(&f)

	// Flush the completed file, even if no promotion is required.
	err = me.t.io.flush(writePath, 0, f.length())
	if err != nil {
		me.logger().Warn(
			"error flushing completed file",
			"file", writePath,
			"err", err,
		)
		err = nil
	}

	if !me.partFiles() {
		return nil
	}

	f.mu.Lock()
	defer f.mu.Unlock()
	f.race++

	renamed, err := me.exclRenameIfExists(f.partFilePath(), f.safeOsPath)
	if err != nil {
		return
	}
	if !renamed {
		return
	}

	err = os.Chmod(f.safeOsPath, filePerm&^0o222)
	if err != nil {
		me.logger().Info(
			"error setting promoted file to read-only",
			"file", f.safeOsPath,
			"err", err,
		)
		err = nil
	}
	return
}
```

The regression test should then expect `a.bin` to be flushed when part files are disabled, while still asserting that no rename occurs:

```go
qt.Assert(t, qt.DeepEquals(ioImpl.flushed, []string{"a.bin"}))
qt.Assert(t, qt.IsFalse(ioImpl.renamed))
```

`MarkComplete` already flushes the correct per-piece extents when `PieceCompletion` is persistent, so this mainly matters for non-persistent or custom completion implementations. Still, preserving the full-file flush seems more consistent with the original intent.

Does this match your understanding and the behavior you had in mind? If so, I can submit a follow-up PR.

PR #1087, opened 2026-08-25
storage: cache classic file writer handles

Aodi:
The classic file backend opened and closed a file for every chunk write. This adds avoidable filesystem overhead on active downloads, particularly on Windows.

This change:

- keeps writable files cached for the lifetime of the backend;
- returns lightweight writers whose `Close` leaves the cached handle open;
- reuses an existing writer handle for storage-internal reads;
- flushes cached handles directly; and
- closes affected handles before rename so part-file promotion remains safe.

The existing default backend selection is unchanged.

Tests cover cached-handle rename and borrowed reads.

Tests:

- `go test -timeout=5m ./...`

### DIFF
--- a/storage/file-io-classic.go
+++ b/storage/file-io-classic.go
@@ -1,36 +1,126 @@
 package storage
 
 import (
+	"errors"
 	"io"
+	"io/fs"
 	"os"
+	"sync"
 )
 
-type classicFileIo struct{}
-
-// Lifetimes of files are scoped to their use, so let's hope everyone is being a good citizen.
-func (me classicFileIo) Close() error {
-	return nil
+type classicFileIo struct {
+	mu      sync.Mutex
+	writers map[string]*os.File
+	closed  bool
 }
 
-func (me classicFileIo) rename(from, to string) error {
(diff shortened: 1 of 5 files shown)


anacrolix:
Just curious why you are depending on the classic IO? Great PR, thanks.

Aodi:
> Just curious why you are depending on the classic IO? Great PR, thanks.

Using classic IO on Windows does not lock files, so it won't prevent files from being modified or deleted. I developed a tool for non-technical people, and I don't want the tool to stop them from deleting their own files – that would confuse them. I know mmap is efficient, and I understand that files shouldn't be deleted/modified during seeding(even mostly they won't do this), but this is a tool for less technical users, and they don't need to worry about these things.

anacrolix:
Ok, doesn't holding the file handles open for classic prevent the files from being removed still?

Aodi:
> Ok, doesn't holding the file handles open for classic prevent the files from being removed still?

The cache handles are only open during download; I close them when the download completes, so seeding won't hold the files. mmap would keep them locked the whole time.

PR #1090, opened 2026-08-26
Try to fix "panic: is zero" when adding pure v2 torrent (#1089)

### LINKED ISSUE #1089: AddTorrentOpt panics (panic: is zero) when adding a pure v2-only torrent, since v1.60.0
## Summary

Adding a torrent with a v2 infohash but no v1 infohash (a real BEP 52 v2-only torrent, not hybrid) panics immediately in `Client.AddTorrentOpt`. Tested on v1.60.0, v1.61.0, and `master` @ 7e8d40966b7c. 

## Repro

```go
package main

import "github.com/anacrolix/torrent"

func main() {
	cfg := torrent.NewDefaultClientConfig()
	cfg.DataDir = "./downloads"

	client, _ := torrent.NewClient(cfg)
	defer client.Close()

	client.AddTorrentFromFile("testdata/bittorrent-v2-test.torrent")
}
```

The repo's own `testdata/bittorrent-v2-test.torrent` (pure v2, no legacy `pieces` field) triggers it — no networking involved, panics synchronously inside `AddTorrentFromFile`.

## Panic (v1.60.0, v1.61.0, `master` @ `7e8d40966b7c`)

```
panic: is zero

goroutine 1 [running]:
github.com/anacrolix/missinggo/v2/panicif.Zero[...](...)
        missinggo/v2@v2.10.0/panicif/panicif.go:112
github.com/anacrolix/torrent.(*Client).AddTorrentOpt(...)
        client.go:1587
github.com/anacrolix/torrent.(*Client).AddTorrentSpec(...)
        client.go:1649
github.com/anacrolix/torrent.(*Client).AddTorrent(...)
        client.go:1761
github.com/anacrolix/torrent.(*Client).AddTorrentFromFile(...)
        client.go:1770
```

## Root cause

`client.go`, `AddTorrentOpt`:

```go
func (cl *Client) AddTorrentOpt(opts AddTorrentOpts) (t *Torrent, new bool) {
	infoHash := opts.InfoHash
	panicif.Zero(infoHash)   // <-- unconditional
	...
	if opts.InfoHashV2.Ok {
		...
	}
```

This asserts the v1 infohash is non-zero unconditionally, before ever checking `opts.InfoHashV2`. A pure v2-only torrent legitimately has an all-zero v1 hash, so this fires every time regardless of a valid `InfoHashV2`. The `if opts.InfoHashV2.Ok` branch right below suggests v2-only was meant to be a valid alternative — the assertion just needs to not fire when a v2 hash is present. Something like `panicif.True(infoHash.IsZero() && !opts.InfoHashV2.Ok)` would do it. Happy to send a PR if that direction looks right.

## Note on pre-v1.60.0

v2-only was already broken before this: v1.59.0/v1.59.1 don't hit this particular panic, but crash differently (nil pointer in webseed request generation, `webseed-requesting.go:372`) after a few MB download. Webseed handling had a lot of work done on it in v1.60.0 though, so that older crash may already be moot — flagging mainly to show v2-only support has not worked across this whole range, not as something to chase separately here.

Related: #175 (general BEP 52 tracking)

oBusk:
Fixes anacrolix/torrent#1089

So I added a test (c4f78539ce7d52550121faed1941f22872f5caf1) using the existing v2 test torrent, to try to catch out issues like this.

Then fixed (309e13b3b6ad539a7ff2ee9e67c961e0db71f601) the main issue, caused by asserting torrent has a hash, ignoring potential v2 torrents.

But once that was done, tests hang forever instead, and I was a bit lost in sauce why this would be. Claude code suggested fixes (https://github.com/oBusk/anacrolix__torrent/compare/fix-1089...oBusk:anacrolix__torrent:fix-v2-torrents), but I'm not sure about them, so leaving them off PR for now.

### DIFF
--- a/client.go
+++ b/client.go
@@ -1584,7 +1584,7 @@ func (cl *Client) AddTorrentInfoHashWithStorage(
 // then this Storage is ignored and the existing torrent returned with `new` set to `false`.
 func (cl *Client) AddTorrentOpt(opts AddTorrentOpts) (t *Torrent, new bool) {
 	infoHash := opts.InfoHash
-	panicif.Zero(infoHash)
+	panicif.True(infoHash.IsZero() && !opts.InfoHashV2.Ok)
 	cl.lock()
 	t, ok := cl.torrentsByShortHash[infoHash]
 	if ok {
--- a/client_test.go
+++ b/client_test.go
@@ -200,6 +200,19 @@ func TestCompletedPieceWrongSize(t *testing.T) {
 	qt.Check(t, qt.IsNil(iotest.TestReader(r, []byte(testutil.GreetingFileContents))))
 }
 
+// https://github.com/anacrolix/torrent/issues/1089
+func TestAddPureV2Torrent(t *testing.T) {
+	cl, err := NewClient(TestingConfig(t))
+	qt.Assert(t, qt.IsNil(err))
+	defer cl.Close()
+	tt, err := cl.AddTorrentFromFile("testdata/bittorrent-v2-test.torrent")
+	qt.Assert(t, qt.IsNil(err))
+	defer tt.Drop()
+	qt.Check(t, qt.IsFalse(tt.infoHash.Ok))
+	qt.Check(t, qt.IsTrue(tt.infoHashV2.Ok))
+	qt.Check(t, qt.Not(qt.DeepEquals(tt.InfoHash(), metainfo.Hash{})))
+}
+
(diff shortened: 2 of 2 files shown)


anacrolix:
I think the tests are correct. Even though they fail I'll merge them.

anacrolix:
Created #1091 from the extra commits.

PR #1091, opened 2026-08-30
Fix v2 torrents

anacrolix:


anacrolix:
@copilot explain the data race cause and fix

anacrolix:
I don't understand, I thought I'd pushed these changes manually but they still show up in the diff.

Issue #1094, opened 2026-08-31
Request stealing produces 11-13% duplicate piece data on high-latency paths with many peers


ozsay:
## Summary

On a high-latency path with a large peer set, `ChunksReadWasted` reaches **one chunk in eight**, and duplicate piece data accounts for **11–13% of every byte received**. All of it comes from request stealing in `applyRequestState`.

This is the same class of problem as #253, and by the standard set there — *"With requestStrategy 3, I'm seeing wastage of less than 1%"* — it is an order of magnitude out.

## Measurements

v1.61.0, three public-tracker torrents totalling ~86 GB, downloaded through a commercial WireGuard VPN exit. `EstablishedConnsPerTorrent` 200, upload capped at 100 kB/s. Each figure is a 480 s window taken after a 300 s ramp, from counter deltas — never sampled rates.

| peers | wire rx | payload delivered | duplicate piece data | `wasted/(useful+wasted)` |
|---|---|---|---|---|
| 175 | 48.91 MB/s | 39.89 MB/s | 11.4% | 12.28% |
| 221 | 54.20 MB/s | 43.05 MB/s | 13.2% | 14.26% |
| 241 | 65.20 MB/s | 52.53 MB/s | 11.9% | 12.83% |
| 193 | 41.61 MB/s | 35.01 MB/s | 8.7% | 9.39% |
| 239 | 66.02 MB/s | 55.12 MB/s | 9.6% | 10.3% |

Transmission 4.1.0 on the same path and the same torrents sits at **4–5%**.

It scales with peer count. At 35 peers on the same setup, duplicate piece data is **1.0%**; the problem only appears once the swarm is large enough that steal candidates are plentiful.

## How this was measured

The torrent-level `ConnStats` counters, read from the `Torrent` rather than summed over live peers — `Peer.modifyRelevantConnStats` writes through, so they survive a disconnect. Summing across live `PeerConn`s instead undercounted delivered bytes by 36% and was biased: reaped peers are the slow ones, which are exactly the peers whose requests get stolen.

Divided against the interface counters they account for every byte:

```
wire rx         - BytesRead         below the app: IP/TCP headers, retransmits, DHT, trackers
BytesRead       - BytesReadData     protocol: handshakes, have, bitfield, PEX
BytesReadData   - BytesReadUseful   piece data delivered twice and discarded
BytesReadUseful - progress          accepted, then failed hash
```

[...]

Issue #253, opened 2018-05-31
Redundancy of loading


YouROK:
When viewing a movie in an Elementum or in my program (similar to an Elementum https://github.com/YouROK/TorrServe), downloading useful data (BytesReadUsefulData) is ~20 times smaller than BytesReadData
Torrent 7.7 GB
In stat:
BytesRead ~9.57 GB
BytesReadData ~9.55 GB
BytesReadUsefulData ~564 MB

Therefore, the download of the network channel is very large

Sorry for my english, this is not my native language

anacrolix:
There's all manner of possible causes for this. Could you provide a way to reproduce the problem?

YouROK:
```
package main
import (
	"log"
	"os"
	"sync"
	"time"

	"github.com/anacrolix/torrent"
)

func main() {
	config := &torrent.Config{
		EstablishedConnsPerTorrent: 150,
		HalfOpenConnsPerTorrent:    97,
		DisableIPv6:                true,
	}

	client, err := torrent.NewClient(config)
	if err != nil {
		log.Fatal(err)
	}

	//Ubuntu
	t, err := client.AddMagnet("magnet:?xt=urn:btih:e4be9e4db876e3e3179778b03e906297be5c8dbe&dn=ubuntu-18.04-desktop-amd64.iso&tr=http%3a%2f%2ftorrent.ubuntu.com%3a6969%2fannounce&tr=http%3a%2f%2fipv6.torrent.ubuntu.com%3a6969%2fannounce")
	if err != nil {
		log.Fatal(err)
	}

	<-t.GotInfo()
	file := t.Files()[0]

	reader := file.NewReader()
	var wa sync.WaitGroup
	wa.Add(1)

	go func() {
		buf := make([]byte, 4*1024*1024)
		for {
			_, err := reader.Read(buf)
			if err != nil {
				break
			}
		}
		wa.Done()
	}()

	go func() {
		cl := t.Closed()
		for {
			select {
			case <-cl:
				return
			default:
				client.WriteStatus(os.Stdout)
			}
			time.Sleep(time.Second)
		}
	}()
	wa.Wait()
}
```

After a while: 185mb useful and 557mb of other
  BytesRead: (int64) 557242945,
  BytesReadData: (int64) 555204608,
  BytesReadUsefulData: (int64) 185155584,
The longer the loading, the greater the difference

Log:
Listen port: 46663
[...]

PR #1095, opened 2026-08-31
Add ClientConfig.StealRequestGrace: a minimum request age before stealing

ozsay:
Fixes most of #1094.

## What this changes

`ClientConfig.StealRequestGrace` — a minimum age a request must reach before another peer is allowed to steal it. Checked last in the steal block, so it costs nothing until a candidate has already passed the queue-depth and recency tests:

```go
// Don't steal a request the current holder has not had time to answer. The tests
// above compare queue depths, which say nothing about whether the block is already
// on the wire; see ClientConfig.StealRequestGrace.
if !t.stealRequestGraceElapsed(req) {
    continue
}
t.cancelRequest(req)
```

The clock is `requestState.when`, which `PeerConn.request` rewrites on every issue, so the grace is **per-holder**: each peer gets one uninterrupted attempt, and a single request cannot migrate more than once per grace period. The "don't steal from the poor" and recency tests are untouched.

**Zero or negative disables it, and zero is the default**, so `NewDefaultClientConfig` behaves exactly as it does today. Nothing changes for anyone who does not opt in.

## Measurements

Same setup as #1094: three public-tracker torrents totalling ~86 GB over a commercial WireGuard VPN exit, `EstablishedConnsPerTorrent` 200, 480 s windows after a 300 s ramp, all figures from counter deltas.

One binary throughout — the only thing differing between arms is the grace — with controls interleaved to bound swarm drift.

| grace | peers | wire MB/s | payload MB/s | duplicate | `wasted/(useful+wasted)` |
|---|---|---|---|---|---|
| 0 | 175 | 48.91 | 39.89 | 11.4% | 12.28% |
| 0 | 221 | 54.20 | 43.05 | 13.2% | 14.26% |
| 0 | 241 | 65.20 | 52.53 | 11.9% | 12.83% |
| 25 ms | 253 | 63.99 | 52.24 | 11.3% | 12.18% |
| 100 ms | 243 | 66.87 | 54.67 | 11.1% | 11.88% |
| **250 ms** | 212 | 69.92 | 59.51 | **7.5%** | **7.99%** |
| **1000 ms** | 213 | 53.14 | 46.64 | **5.1%** | **5.52%** |

A second sweep after the test machine's disk and RAM changed, so these are bracketed by their own controls and are not comparable to the table above:

| grace | peers | wire MB/s | payload MB/s | duplicate | `wasted/(useful+wasted)` |
|---|---|---|---|---|---|
| 0 | 193 | 41.61 | 35.01 | 8.7% | 9.39% |
| **2000 ms** | 228 | 50.43 | 45.36 | **2.6%** | **2.8%** |
| 0 | 239 | 66.02 | 55.12 | 9.6% | 10.3% |

### It is a threshold, not a gradient

25 ms and 100 ms are indistinguishable from control. The 100 ms arm and its nearest control are the closest-matched pair in the set — 243 vs 241 peers, 66.87 vs 65.20 MB/s — and differ by 0.8 points, inside the control-to-control spread.

Nothing happens until somewhere between 100 and 250 ms, then it bites. That is consistent with the grace needing to exceed the path's round-trip before a cancel can prevent anything, and it argues against the alternative reading that this works by limiting churn — churn limiting would have paid off smoothly from 25 ms upward.

Measured against each arm's own contemporaneous control, the series is monotone and has not flattened:

| grace | fraction of its control's duplicates |
|---|---|
| 250 ms | 0.61 |
| 1000 ms | 0.42 |
| 2000 ms | 0.28 |

### It is not swarm drift

Five control arms across two sweeps: 11.4 / 13.2 / 11.9 / 8.7 / 9.6%. Two pairs are matched closely enough to read directly:

| | peers | wire | duplicate |
|---|---|---|---|
| control | 221 | 54.20 | 13.2% |
| 1000 ms | 213 | 53.14 | **5.1%** |

Essentially one operating point, 2.6x apart. And the 2000 ms arm ran **busier than both of its controls** — 228 peers against 193 and 239, 50.4 MB/s against 41.6 — which by the peer-count scaling in #1094 predicts *more* duplication, not less. It produced 3.5x less, so drift can only be understating the effect.

### No regression

Payload never fell below the controls' range and peer count never dropped, so the grace is not suppressing connections or starving the request pipeline:

| | controls | treatments |
|---|---|---|
| payload MB/s | 35.0–55.1 | 45.4–59.5 |
| peers at end | 175–241 | 212–253 |

## What this does not do

**It does not reach the under-1% of #253.** At 2000 ms duplicate piece data is 2.6% — below Transmission's 4–5% on the same path for the first time, but the remaining ~2.6% is not explained by this mechanism.

**A constant is the wrong shape.** The threshold is a property of the path's round-trip time, and 2000 ms is a generous guess at one exit's. A per-peer RTT estimate is what this actually wants; there is none in the library today, so the constant is what a first version can do. If a peer RTT ever lands, this should read from it.

**The tail is unmeasured.** Every window here is 480 s taken mid-download. With no endgame mode, a request parked on a slow peer near completion cannot be re-issued for the length of the grace. That is the reason the default is off, and the first thing to turn down if last-piece latency looks wrong.

If the request-age idea is sound but the interface is not, I am happy to rework it — an internal constant, a bool, or keying it off something existing rather than a new `ClientConfig` field.

## Tests

`steal-request-grace_test.go` covers the predicate (disabled, younger than, at, and older than the grace), that the grace is per-holder rather than per-request lifetime, and the `NewDefaultClientConfig` plumbing including a malformed environment value. The existing suite passes unchanged with the grace at 0, 250 ms and 1 s — the 1 s run is the useful one, since it shows an aggressive grace does not stall the end-to-end download tests.

### DIFF
--- a/config.go
+++ b/config.go
@@ -92,6 +92,26 @@ type ClientConfig struct {
 	DownloadRateLimiter *rate.Limiter
 	// Maximum unverified bytes across all torrents. Not used if zero.
 	MaxUnverifiedBytes int64
+	// Minimum age an outstanding request must reach before another peer is allowed to steal it.
+	// Zero or negative disables the check, which is the historical behaviour.
+	//
+	// Stealing (see PeerConn.applyRequestState) hands a request to a peer that looks better, and
+	// its trigger is a one-request difference in queue depth. Across a large peer set that
+	// difference is noise: queue depths move by one every time a chunk lands, so without a floor
+	// the same request can migrate several times inside a single round-trip. Each migration is a
+	// Cancel on the wire, and a Cancel only prevents a transfer if it reaches the peer before the
+	// peer has served the request. Otherwise the block arrives regardless and is thrown away as
+	// ConnStats.ChunksReadWasted, while the stealer requests the same block again.
+	//
+	// A request younger than a round-trip has not yet had the opportunity to be answered, so its
+	// holder has not been shown to be failing and there is no information in moving it. Requiring
+	// an age gives each holder one uninterrupted attempt, and because the age is reset every time
+	// the request is issued (PeerConn.request), it also caps how often a single request can move.
+	//
+	// Latency is what makes this matter: the data already committed to the wire when a Cancel is
+	// sent scales with round-trip time, so on a LAN the cost of a steal is near zero and on a
+	// tunnelled path it is not.
+	StealRequestGrace time.Duration
 
 	// User-provided Client peer ID. If not present, one is generated automatically.
 	PeerID string
@@ -255,6 +275,7 @@ func NewDefaultClientConfig() *ClientConfig {
 		DialForPeerConns:       true,
 		AcceptPeerConnections:  true,
 		MaxUnverifiedBytes:     64 << 20,
+		StealRequestGrace:      initDurationFromEnv(stealRequestGraceEnvKey, defaultStealRequestGrace),
 		DialRateLimiter:        rate.NewLimiter(10, 10),
 		PieceHashersPerTorrent: 2,
 	}
--- a/env.go
+++ b/env.go
@@ -3,6 +3,7 @@ package torrent
 import (
 	"os"
 	"strconv"
+	"time"
 
 	"github.com/anacrolix/missinggo/v2/panicif"
 	"golang.org/x/exp/constraints"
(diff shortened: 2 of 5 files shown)


anacrolix:
Very good. What do you think of making 250ms the default?

ozsay:
Testing direct downloads without a VPN yielded no noticeable improvement in waste percentage over Transmission. I haven't completed a full torrent download to quantify how grace impacts the end of a download. 

This feature isn't battle-tested enough to adopt as the default. Many variables that are not the engine itself may have been affected the results (VPN load, swarm...).

I believe that there are many unknowns that indicate that the grace should be dynamic according to swarm health, RTT etc.

PR #1088, opened 2026-08-25
storage: add buffered piece completion wrapper

Aodi:
Persisting every completed piece in its own Bolt transaction adds synchronous database work to the download hot path. This PR adds the buffering infrastructure needed to avoid that work, without enabling the buffered wrapper for Bolt by default yet.

This change:

- adds optional batch/checkpoint interfaces for piece completion implementations;
- adds `PieceCompletionChange` for multi-update persistence;
- implements `SetBatch` for Bolt piece completion;
- introduces `bufferedPieceCompletion`, an in-memory overlay wrapper that defers `complete=true` writes until checkpoint or close; and
- keeps incomplete and unknown states persisted immediately so stale completion data cannot survive invalidation.

The actual Bolt wrapping is intentionally left for a follow-up checkpoint batching PR, where the file flush and `PieceCompletionCheckpointer` call sites are introduced together. That keeps this PR from changing the current persistence behavior on its own.

Tests cover deferred completion, immediate false and unknown invalidation, batch checkpoint persistence, close-time flushing, and persistent-state forwarding.

Tests:

- `go test ./storage`

PR #1103, opened 2026-09-07
storage: batch file and completion checkpoints

Aodi:
Depends on #1088. Also builds on #1086, which has already been merged.

Persistent piece completion currently flushes file data for every completed piece before recording completion. This creates repeated synchronous flush and database work on the download hot path.

This change:

- enables the buffered wrapper for Bolt piece completion;
- queues completed pieces and their files when the completion store supports checkpoints;
- flushes queued files and checkpoints queued completion records at a one-second or 128 MiB threshold;
- flushes and checkpoints pending work when the torrent closes; and
- snapshots and restores pending state around I/O so concurrent queueing remains safe and failed checkpoints can be retried.

The buffered wrapper and batch/checkpoint interfaces are introduced in #1088. Bolt is wired through the wrapper here because this PR also adds the file flush and `PieceCompletionCheckpointer` call sites needed to preserve persistence behavior.

Tests cover deferred flush, close-time persistence, and concurrent queueing.

Tests:

- `go test ./storage`

### DIFF
--- a/storage/bolt-piece-completion.go
+++ b/storage/bolt-piece-completion.go
@@ -43,7 +43,7 @@ func NewBoltPieceCompletion(dir string) (ret PieceCompletion, err error) {
 		return
 	}
 	db.NoSync = true
-	ret = &boltPieceCompletion{db}
+	ret = newBufferedPieceCompletion(&boltPieceCompletion{db})
 	return
 }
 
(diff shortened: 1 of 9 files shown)


anacrolix:
@copilot resolve the merge conflicts in this pull request

anacrolix:
https://github.com/anacrolix/torrent/pull/1088 merged

